PR #73, opened 2026-07-23
Sentry-compatible error alerting for production 500s (C3)

ibuilder:
First Tier-2 enterprise item. Closes the "500s page no one" SLA blocker (C3) — the DB `error_log` is a record, not an alarm. Adds Sentry-compatible external error alerting, **server-side only, no-op until a DSN is set**, so it lands safely with no provider chosen (Sentry SaaS, self-hosted Sentry, and GlitchTip all use the same DSN-style integration).

## Changes
- **`services/api/src/aec_api/sentry.py`** (new) — env-gated `init()`, `capture_exception()`, and a `before_send` scrubber. No-op until `AEC_SENTRY_DSN`/`SENTRY_DSN` is set; `_ENABLED` flips true only after a successful init so the unset path provably never touches the SDK/network. `traces_sample_rate` defaults to **0** (error reporting only — tracing is a separate Tier-2 item). Everything fails open: init/capture failures are logged and swallowed, never block boot or mask a 500.
- **`main.py`** — `sentry.init()` early in `lifespan` (before init_db, so startup failures are reported); `sentry.capture_exception(exc, request_id=rid)` inside the existing global 500 handler. The handler catches before Sentry's auto-capture would, so this is explicit. The DB `error_log` write + clean 500 + `X-Request-ID` response are unchanged — Sentry is additive.
- **`requirements.in`** — add `sentry-sdk[fastapi]>=2.0`.
- **`requirements.lock`** — regenerated via `pip-compile --generate-hashes --allow-unsafe` under Python 3.12 (matches `lockfile.yml`); adds `sentry-sdk[fastapi]==2.66.1` + transitive deps, no other pins changed.
- **`.env.example`** — documents `AEC_SENTRY_DSN`, `AEC_SENTRY_ENVIRONMENT` (default `AEC_ENV`), `AEC_SENTRY_RELEASE`, `AEC_SENTRY_TRACES_SAMPLE_RATE`, with an operator note.
- **`test_sentry.py`** — DSN-unset no-op; DSN-set init + capture + request-id tag; capture fail-open; `before_send` scrubs authorization/cookie/x-api-key headers + request body. All pass; `test_errorlog.py` unchanged (no regression).

## Privacy / safety
`send_default_pii=False`; explicit `before_send` scrubs `authorization`/`cookie`/`x-api-key` headers and replaces request `data` with `[scrubbed]` (and drops the whole request payload if scrubbing itself fails). Each event is tagged with the existing request-id for correlation with the JSON access log and the `error_log` row.

## To enable (operator)
Set `AEC_SENTRY_DSN` and create alert rules in your Sentry/GlitchTip instance. Until then, behavior is identical to today.

\`\`\`task
file:services/api/src/aec_api/sentry.py
file:services/api/src/aec_api/main.py
file:services/api/requirements.in
file:services/api/requirements.lock
file:.env.example
\`\`\`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -89,6 +89,16 @@ AEC_LOGIN_WINDOW_SEC=300                          # ...within this sliding windo
 # this value as the bearer token. Unset = the whole /scim/v2 surface returns 503 (disabled).
 # AEC_SCIM_TOKEN=
 
+# --- Error alerting (optional) — Sentry / GlitchTip ---
+# The DB error_log records what broke but alarms no one. Set a DSN to page on-call on production 500s
+# (dedup + release tracking). Server-side only; UNSET = no-op (nothing initializes, no network). Sentry
+# and self-hosted GlitchTip both take the same DSN-style value. After setting the DSN, create alert
+# rules in your Sentry/GlitchTip project. Error reporting only — tracing stays off by default.
+# AEC_SENTRY_DSN=https://<key>@o0.ingest.sentry.io/0   # or your GlitchTip DSN; SENTRY_DSN also honored
+# AEC_SENTRY_ENVIRONMENT=production                    # defaults to AEC_ENV
+# AEC_SENTRY_RELEASE=massing@0.1.0                      # optional; tags events with a release for regressions
+# AEC_SENTRY_TRACES_SAMPLE_RATE=0                       # keep 0 — performance tracing is a separate concern
+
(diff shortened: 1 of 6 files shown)


ibuilder:
Superseded — the Sentry error-alerting work was already contained in #75 (chore/otel-tracing), which landed in v0.3.587. Closing as redundant.

PR #74, opened 2026-07-23
Alembic DB migrations — versioned baseline + drift guard (C1)

ibuilder:
Second Tier-2 enterprise item. Closes the C1 blocker: today the schema is `Base.metadata.create_all()` + additive-only `_ensure_columns`/`_ensure_indexes` (db.py) — no versioning, no rollback, no audit trail. This PR introduces **Alembic as the migration framework with a baseline-stamp, no runtime cutover**.

**Stacked on #73** (chore/sentry-error-alerting) to avoid a second independent `requirements.lock` conflict — both PRs add Python deps. Once #73 merges, this rebases to `main` cleanly.

## What lands
- **`alembic.ini` + `migrations/env.py`** — env.py assembles the **complete** metadata exactly the way the app does at startup: the static `aec_api.models` plus the 130 config-driven dynamic `mod_<key>` tables (registered into `Base.metadata` by `modules_registry.load_registry()`). Autogenerate / `alembic check` therefore see exactly the schema the running app manages. DB URL comes from `DATABASE_URL` (never hardcoded).
- **One baseline revision** generated against an **empty** DB (not a `create_all`-built one, which would yield an empty migration) — it CREATEs all **158 tables**. Validated: `upgrade head` succeeds on a fresh DB and `downgrade base` cleans up.

## Drift audit (key finding)
Diffed a `create_all` + `_ensure_*` DB against an Alembic-upgraded DB: **856 objects each, zero difference** — `_ensure_columns`/`_ensure_indexes` add nothing beyond model metadata. The **one** real drift is the Postgres-only full-text **GIN** indexes (`ix_mod_<key>_fts`, raw `to_tsvector(...)` expression DDL from `modules.ensure_fts_indexes`, invisible to autogenerate). Reconciled into the baseline by reusing the same DDL builder (`modules_search.index_ddl`), and `env.py` excludes them from `alembic check` so the drift guard stays clean.

## No runtime cutover
`create_all` / `_ensure_columns` / `_ensure_indexes` remain the active startup path, untouched. Nothing runs `alembic upgrade` at boot. The Dockerfile only COPYs `alembic.ini` + `migrations/` so operators can run `alembic stamp/upgrade head` from the container. Replacing the additive path with migration-driven schema management is a **separate follow-up PR** (the cutover).

## Drift guard (CI + test)
- **`test_alembic_migrations.py`** (in the test gate) — `alembic upgrade head` on a fresh DB creates all 158 tables; `alembic check` reports clean (no model-vs-migration diff). Passes on SQLite.
- **`db-migrations.yml`** workflow — full Postgres guard: `upgrade head` → `alembic check` → diff a runtime-built DB (`create_all` + `_ensure_*`) vs an Alembic-upgraded DB (`pg_indexes` + column comparison) to catch any `_ensure_indexes` drift. (Not run locally — no Postgres in the sandbox; SQLite parity equivalent verified.)

## Operator docs
`migrations/README.md`: existing DBs → back up, verify schema, then `alembic stamp head` (marks them at the baseline WITHOUT running it); fresh DBs → `alembic upgrade head`. States Alembic is now the source of truth; future schema changes via `alembic revision`.

## Lock
`requirements.lock` regenerated with the exact CI command (`pip-compile --generate-hashes --allow-unsafe`, Python 3.12) — minimal diff: only `alembic==1.18.5` + `mako==1.3.12`.

\`\`\`task
file:services/api/alembic.ini
file:services/api/migrations/env.py
file:services/api/requirements.in
file:services/api/requirements.lock
file:.github/workflows/db-migrations.yml
\`\`\`


ibuilder:
Landed in v0.3.587 (merged onto current main, full suite re-verified 316/316; deps in requirements.lock so CI installs them).

PR #94, opened 2026-07-29
R23-DIGEST — a model summary you can subtract

ibuilder:
Ring item **R23-DIGEST** (engineering upgrade ring): *"a deterministic multi-scale model digest as compact JSON. Immediate non-AI value as a diffable change-detection snapshot between IFC versions."*

## What it does

Two IFCs land a fortnight apart and "what changed?" is answered by opening both and looking. This turns it into arithmetic: `project → storey → class → element`, where every node is hashed over its own content **plus its children's hashes**. That Merkle shape is the point — comparing two publishes is one string comparison, and when it differs you descend only into the storeys whose hashes differ. You never walk the elements of a storey that did not change.

## Three decisions worth review

**Systems and zones are axes, not levels.** The ring entry names the scales as `project → storey → zone → system → element`. That hierarchy does not exist in IFC — zones group spaces, systems group distribution elements, both cut across storeys, and an element can be in several. Built as one spatial spine that owns elements plus two grouping axes that reference them by GlobalId, so a duct running three floors is counted once and appears in all four places.

**Determinism is a claim, so it is a test.** A digest is worthless if an unchanged building hashes differently, and every way that happens is silent: a path, a stat, a timestamp, ifcopenshell's per-file STEP id inside a pset, float noise in a volume. `test_model_digest` digests a copy **at a different path** and asserts byte-identical JSON — the assertion that catches a path or stat leak, which comparing a file to itself never would.

**A digest states its configuration, and the diff enforces it.** Geometry-derived and pset-derived quantities disagree for reasons that have nothing to do with the building. `diff()` refuses across modes, rounding, versions or prune depths with a stated reason rather than reporting a configuration change as building change.

The route deliberately does **not** inherit the baseline's measurement mode. Mirroring it would be more convenient, and it would let a string in caller-supplied JSON decide whether a request meshes every element in the model — while disarming the guard, since the two sides would then always agree.

## Routes

```
GET  /projects/{pid}/digest?scale=&geometry=&summary=
POST /projects/{pid}/digest/diff
```

Stateless by design, not by shortcut: there is no version table behind `source_ifc`, a storey-pruned digest is a few kilobytes, and the caller already has somewhere to keep it. A baselines table would mean a migration and a retention policy for data whose whole purpose is to be compared once.

## Verification

- 35 engine checks + 25 HTTP checks green
- `route_authz`, `reachable`, `engine_routes`, `file_sizes`, `import_cycles` pass

## Not included, deliberately

Version bump, CHANGELOG and roadmap entries — those are the files that conflict. Left for merge time so this lands alongside concurrent work.


ibuilder:
Follow-up commit: the two endpoints moved out of `analysis.py` into their own `routers/digest.py`.

Two reasons. `analysis.py` describes itself as clash detection and IDS validation, and a model digest is neither. And concurrent work is actively editing that file — the two changes were ~700 lines apart and merged clean, but "clean today" isn't worth relying on when the alternative is not sharing the file at all. **This branch no longer touches `analysis.py`** (`git diff main -- analysis.py` is empty).

No behaviour change: same two paths, same handlers. `digest_route` (25 checks) re-verified, ruff clean, route-authz / reachable / route-order pass. CI is re-running for the new commit; the previous one was fully green including the 17-minute API test gate.

ibuilder:
Follow-up: merged current `main` in, and the merge exposed a real bug in the digest that had been passing by luck.

**The bug.** `_element_record` deliberately excludes the GlobalId from an element's hash, so a re-export that renumbers every id still compares. But `_class_nodes` sorted those records **by** GlobalId and `_node_hash` folded children in list order — so the parent hash depended on something the child hashes did not. Two models holding identical elements hashed differently whenever the random GUID order happened to differ.

That is not a difference in the building, and it is the worst kind for this feature specifically: it reports a storey as changed when nothing moved, which makes the "descend only into storeys whose hash differs" optimisation unsound.

**Fixed two ways**, each correct on its own terms: element records now order by `(hash, guid)` — content first, GUID only as a tiebreak so output stays stable within one file — and `_node_hash` folds **sorted** child hashes, so a node's hash describes the *set* of its children rather than their emission order. A storey is a collection of elements; iteration order should not be able to say two identical ones differ.

The invariant is now pinned in its own right rather than resting inside the sibling-isolation check: the two separately-authored fixtures are asserted to have **disjoint GlobalIds and matching element-hash sets**, so a future change that reintroduces GUID-dependence fails on the invariant rather than showing up as an intermittent failure downstream.

Ran `test_model_digest` three times (it was order-flaky, so once proves little) — green each time. `digest_route` green, ruff clean.

Merge conflict was one line in the `run_tests` manifest (main gained `test_sheet_recover` + `test_firm_standards` on the same long line this branch appends to); both sides kept. `analysis.py` auto-merged and is byte-identical to main.

PR #95, opened 2026-07-29
R22-NOTICE-CLOCK — the contract's clock, counted

ibuilder:
Ring item **R22-NOTICE-CLOCK** (competitive gap ring, Tier 2): *"contractual notice clocks / time-bar tracking. Detect a triggering event in a daily log or RFI, start the contract's notice period, draft the notice. Highest dollar-per-line-of-code feature in construction administration; we already hold the contract calendar and the daily record."*

## What it does

A contract gives you a fixed number of days to notify the other side of an event before you lose the right to be paid or given time for it. The events are already in this platform — a stoppage in a daily report, an unforeseen condition on a change event, an RFI flagged cost-impacting — and the clock starts the day they land. Nothing counted them.

`notice_clock.py` carries the standard notice provisions of seven contract families (AIA A201-2017, FIDIC Red Book 2017 and 1999, NEC4 ECC, ConsensusDocs 200, EJCDC C-700, FAR) with their citations, detects triggering events from `daily_report` / `change_event` / `rfi`, and returns a register sorted expired-first.

## Please review the refusals, not the arithmetic

This is a feature whose failure mode is a **confident wrong date**. Nothing throws, nothing looks odd, and somebody relies on it. So most of the design is in what it declines to do, and each of those had a tempting sensible-looking fallback:

- **It will not pick a contract for you.** No family adopted → an empty register and a reason. A deadline computed under a contract the project is not on is not a conservative approximation of the right answer; it is a different answer that looks exactly like it. Adoption lives on `prime_contract.notice_family`.

- **It will not substitute the occurrence date for the awareness date.** FIDIC and NEC run from awareness. Falling back to occurrence *looks* conservative and is the opposite: an event that occurred forty days ago and was discovered yesterday reads as time-barred while three weeks remain, and a claimant told the bar has fallen abandons a live claim. It reports `indeterminate` and names the missing field. `change_event` gained `occurred_on` + `became_aware_on` so there is somewhere to put the answer.

- **It will not invent a period for a clause that fixes none.** FAR 52.236-2 says "promptly, and before the conditions are disturbed" — that returns `no_fixed_period` with the obligation stated, not a plausible number of days.

- **It will not pretend the calendar is universal.** Calendar days, business days and NEC's weeks are three different counts; business days without a holiday list are short one day per holiday (warned, *with the direction of the error*); and the weekend is Fri/Sat across much of the Gulf, so it is an input and is reported back with every result.

## Scope boundary

`consequence` states what the contract **text** says — `express_time_bar` / `condition_precedent` / `contractual_requirement` — and stops there. Whether late notice actually forfeits a claim turns on the wording, the governing law and the forum. This is a diary, not advice, and every response carries the instruction to verify the provision against the executed contract. **The clause table is a starting point to check, not a statement of what any particular contract says** — an amended sub-clause changes the number and this library cannot know it was amended.

## Routes

```
GET  /projects/{pid}/notices/clauses
GET  /projects/{pid}/notices          (?family= override is labelled a what-if)
POST /projects/{pid}/notices/draft
```

## Verification

- 74 engine checks + 36 HTTP checks green
- The module select options and `FAMILY_BY_LABEL` are asserted **set-equal both ways** — two tables encoding one vocabulary drift silently, and the failure mode is a register that goes quietly empty on a project that *had* adopted a family
- `modules`, `module_config`, `module_schema`, `module_rooms`, `route_authz`, `reachable` pass

## Not included, deliberately


ibuilder:
**Security audit finding — latent, now fixed.** A weekend covering all seven days was accepted.

`parse_calendar` validated that each token was a real weekday, but never that any *working day survived*:

```
parse_calendar("Mon,Tue,Wed,Thu,Fri,Sat,Sun") -> weekend=(0,1,2,3,4,5,6)
is_working across a full week: all False
```

`add_period(basis="business")` decrements only when it finds a working day, so with none it walks to `date.max` — ~2.9 million iterations, then an `OverflowError` raised from date arithmetic rather than from anything that knows what went wrong. Not caught on that path, so it would surface as a **500**. The symptom is a slow endpoint; the cause is a bad calendar.

**Latent, and worth being precise:** no clause in the library uses `basis="business"` (count is 0), so `add_period` is only reached with `calendar` or `weeks` today. But `weekend` **is** caller-supplied, so the malformed calendar is constructible right now — it just has nothing to feed.

### Fixed at both layers, because they fail differently

- `parse_calendar` rejects it at the point the mistake is made, naming the consequence — where the caller can still see the field they got wrong.
- `add_period` also refuses before entering the loop. `Calendar` is a plain dataclass, constructible directly without going through `parse_calendar`, and **a `while` whose termination depends on data should not trust the data.**

### Why fix something unreachable

`BASIS` already advertises `"business"`, the branch is written and documented, and the Gulf-weekend reasoning in the module docstring is exactly the thinking that leads someone to add the first business-basis clause. The feature that makes it live and the feature that makes it plausible are the same feature.

### Tests

The raise at parse time; the raise inside `add_period` for a directly-built `Calendar`; a 422 over the wire with the consequence in the body; and the legitimate extreme — a **six**-day weekend, one working day, terminates and returns the right date. That last one guards the natural over-correction (`len(days) > 5`), which would break the Gulf case the generality exists for.

**77 engine checks + 38 HTTP checks green; ruff clean.**

---

The same pass audited #99 and found it clean, re-deriving the arithmetic rather than reading it — including the line that justifies that module better than my own PR description did: on the sample deal the hurdle clears **1.00 conditional on approval, 0.58 unconditional**. Reporting both with the unconditional as the headline is the whole point of separating the branches.

PR #96, opened 2026-07-29
R23-PREFAB-KIT — a kit is a frozen list, not a live query

ibuilder:
Ring item **R23-PREFAB-KIT** (engineering upgrade ring, Tier 3): *"a prefab kit is a `query_dsl.select()` scope + BOM + pull-plan task + delivery date. A join across spines we already have, not a new engine."*

## The join, and the conflict inside it

The ring entry is right that this is a join: a kit is a `query_dsl` scope + a BOM rolled up from those elements + the pull-plan task it installs under + the delivery that has to land before that task starts. Nobody could previously answer *"is the L3 duct kit ready to release, and will it be on site before the crew needs it?"* without opening four screens and comparing by eye.

But the join has a conflict at its centre, and it is why this needed care. **A `query_dsl` selector is live** — it re-evaluates against whatever the model says today. **A kit is fabricated off-site against a scope decided weeks ago.** Resolve that the convenient way (always re-run the selector) and a model revision silently changes what the shop is building after the steel is cut.

So a released kit **freezes its GlobalId list**. The selector becomes the record of how that list was arrived at rather than the definition of it, and `drift()` reports the difference between the frozen scope and what the selector matches now.

**Both directions are reported separately**, because they are different defects: an element that has *entered* scope is work nobody priced; one that has *left* it is a design change the shop has not been told about and may already have built. Netting them to "still 2 elements" would hide both. Same discipline as an issued drawing — the revision cloud is the point.

## What falls out of taking that seriously

- `POST .../freeze` **refuses** an empty, erroring or truncated selector. A released kit with nothing frozen is the exact failure this mechanism exists to prevent, and would look identical to a correct one on every screen.
- The `release` transition requires `frozen_guids` + `released_on`, so the state the shop works from cannot be reached without a written scope.
- A frozen GlobalId the model no longer contains is reported `unresolved`, not skipped.
- **The BOM reads quantities from `qtos`, where the index actually nests them.** Reading them as flat per-element keys — the obvious first guess, and what my first draft did — yields a BOM of zeroes indistinguishable from a measured result. Declared quantities only, no geometry fallback: a shop document should not mix what the model asserts with what we derived by meshing a solid. Elements with no `Qto_*` contribute their count and are reported as `unquantified` rather than counted as zero.
- The register sorts by **blocker severity**, not by date: a drifted kit outranks a merely late one, because a late kit is a known problem and a drifted one is an unknown wrong one.

## Routes + module

New `prefab_kit` module (133 total).

```
GET  /projects/{pid}/prefab/kits
GET  /projects/{pid}/prefab/kits/{rid}
POST /projects/{pid}/prefab/kits/{rid}/freeze
```

## Verification

- 48 engine checks + 31 HTTP checks green
- The HTTP set drives the real loop: **freeze → release → change the model underneath → assert the change is reported as drift rather than absorbed**
- `module_config`, `module_schema`, `module_rooms`, `route_authz`, `route_order`, `reachable`, `modules` pass

## Not included, deliberately


ibuilder:
CI caught a real defect here, now fixed: a new `module.json` creates a `mod_<key>` table at runtime, so `alembic check` reported drift and failed the **alembic upgrade + check + runtime-parity (Postgres)** job. Every local test passed, because SQLite builds tables from the models.

Added the revision for `mod_prefab_kit` and its five indexes, chained to the current head.

It includes the Postgres-only FTS GIN index, which looks like boilerplate and isn't: the baseline migration only indexes tables that existed at *its* point in the chain, so a post-baseline module that skips it silently gets no full-text search on Postgres while passing every SQLite test. The runtime-parity half of the same job is what catches that.

Verified locally — `test_alembic_migrations`: upgrade head builds 162 tables, `alembic check` reports no new upgrade operations.

PR #97, opened 2026-07-29
R22-CLASSIFY-AI — a coverage number that counts only what the model says

ibuilder:
Ring item **R22-CLASSIFY-AI** (competitive gap ring, Tier 2, ⭐): *"assisted classification of imported IFC. We have a canonical discipline tree and a rule that a Uniclass code is never guessed — correct, and it means a client's unclassified model (which is nearly every real model) gets nothing."*

## Checking the premise turned up the more interesting half

`classification.classify(ifc_class, system)` returns a `(code, title)` tuple for **any** input. When the class is not in the system's map it returns the **default bucket** — MasterFormat `01 00 00 General Requirements`, NRM1 `0 Facilitating works` — and the return type cannot express the difference.

A model we author is full of classes that *are* in the map, so this never shows. An imported model is full of `IfcBuildingElementProxy`, and every one of those elements prices as General Requirements while the estimate reports a complete takeoff. Nothing is missing, nothing errors, and the number is wrong.

That is the same shape as the `get_area` defect: a fault that only fires on one *provenance* of model, so the everyday path never catches it. There the provenance was models we authored; here it is **imported**, which is nearly every real client model.

## So the first deliverable is a coverage figure that can be believed

`classified_pct` counts **only what the model declares** via `IfcRelAssociatesClassification`. Derived codes are proposals and never raise it — a percentage that included our own guesses would measure the guessing, not the model. **0% on a client import is the true starting point of a cost exercise**, and more useful than the implicit 100% you get today.

## Second: proposals that carry their evidence

Four ranked bases, each saying what kind of claim it is:

| basis | what it means |
|---|---|
| `declared` | the model asserts it. Not a proposal. |
| `keyword` | the element's **own name** determines the code ("CMU Partition" → unit masonry). Specific to this element. |
| `class_map` | the IFC class is in the published map. True of the **class**, so identical for every element of that class — *including the ones it is wrong for*. NRM1 maps `IfcWall` to **external walls**; an interior wall exported as plain `IfcWall` gets the external code, and nothing about the derivation knows that. Low confidence, and the evidence string says so. |
| `unmapped` | no entry. **The case the current code hides** — reported as a gap, with the bucket it would otherwise have been given named in the evidence. |

**Uniclass and OmniClass are refused outright**, with the standing reason. The tables here are published IFC-class maps, which is a different thing from inferring a Uniclass reference from a name. A refusal with a reason, not an empty result — an empty result reads as "nothing to classify".

**Nothing is written by reading it.** `/classify/plan` turns an *explicit* list of accepted GUIDs into ordinary `set_classification` recipe steps for the caller to apply. A classification that appeared on elements because somebody opened a report is indistinguishable, a week later, from one a quantity surveyor decided. Anything unapplicable comes back in `skipped` **with a reason** rather than being dropped.

## A defect in my own first draft, worth recording

The system-name matcher did substring comparison in both directions, so the display name `"CSI MasterFormat (US/CA)"` contributed the fragment `"us"` and a model declaring **"AusSpec" matched MasterFormat**. A wrong system match is worse than no match — it reports an element as classified under a standard it never was. Now an exact match against an explicit alias list, asserted in both directions.

## Routes

```
GET  /projects/{pid}/classify/coverage?system=
GET  /projects/{pid}/classify/proposals?system=&basis=&limit=
POST /projects/{pid}/classify/plan
```

## Verification

- 47 engine checks + 30 HTTP checks green
- The HTTP set drives the full round trip against a real IFC: **propose → apply through `/edit/batch` → re-read → assert the element now reads as `declared`** — i.e. that what this proposes and what the platform writes are the same thing, read back by a path that did not write them. That round trip is also what caught the endpoint being `/edit/batch`, not the `/edit-batch` my first draft cited in its own note.
- `route_authz` (683 routes), `reachable`, `route_order`, `classification`, `engines` pass
- No new module and no new table, so no Alembic revision is needed here

## Not included, deliberately

### DIFF
--- a/services/api/run_tests.py
+++ b/services/api/run_tests.py
@@ -27,6 +27,8 @@ TESTS = ["test_proforma", "test_cost", "test_modules", "test_dashboard",
          "test_bundle", "test_desktop", "test_localmode", "test_project_budget", "test_rvt_bridge",
          "test_bcf", "test_engines", "test_edge_cases", "test_opendata", "test_financials", "test_money", "test_leasemgmt", "test_changeorders",
          "test_migrate", "test_alembic_migrations", "test_appraisal", "test_marketing", "test_workflow_gate", "test_due_feed", "test_directory",
+         # R22-CLASSIFY-AI — classification coverage + code proposals:
+         "test_classify_assist", "test_classify_route",
          "test_ask", "test_verification", "test_webhooks", "test_operate_capital", "test_payroll_drawings", "test_assistant_itb", "test_construction_depth", "test_distribution", "test_e57", "test_empty_project", "test_metrics", "test_metrics_auth", "test_licensing", "test_revit_bridge", "test_precon", "test_specs", "test_feasibility", "test_clash_import", "test_clash_intel", "test_layout", "test_loads", "test_verified_progress", "test_element_records", "test_securities_bridge", "test_imports", "test_search_alerts", "test_attachments",
          # previously not wired into the gate (glob would have caught these) — now covered:
          "test_analytics", "test_discipline", "test_gbxml", "test_review", "test_interop",
--- a/services/api/src/aec_api/classify_assist.py
+++ b/services/api/src/aec_api/classify_assist.py
@@ -1,0 +1,344 @@
+"""R22-CLASSIFY-AI — cost/discipline classification for a model somebody else authored.
+
+**The finding this is built around.** `classification.classify(ifc_class, system)` returns a
+`(code, title)` tuple for any input. When the class is not in the system's map it returns the
+system's **default bucket** — MasterFormat `01 00 00 General Requirements`, NRM1 `0 Facilitating
+works` — and the return type cannot express the difference. A model we authored is full of classes
+that *are* in the map, so this never shows. An imported model is full of `IfcBuildingElementProxy`,
+and every one of those elements prices as General Requirements while the estimate reports a complete
+takeoff. Nothing is missing, nothing errors, and the number is wrong.
+
+That is the same shape as the `get_area` defect: a fault that only fires on a particular provenance
+of model, so the everyday path never catches it. Here the provenance is *imported*, which is nearly
+every real client model.
+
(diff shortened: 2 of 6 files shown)


ibuilder:
**Coverage denominator fix.** Raised as a nit by the security-audit session: `IfcOpeningElement` is an `IfcElement` subtype, so voids landed in `total` and could never carry a code.

Checked the premise before acting, and it was right in substance but not in location. `ifc_loader.physical_elements` does **not** filter openings — `qto.py` skips them at **both** its call sites (lines 241 and 325). So the convention here is "openings are not priced", expressed at the call site rather than in the shared helper.

That changes the verdict from *comment either way* to *fix*. Classification feeds QTO and cost, so a denominator containing openings makes the coverage figure disagree with the takeoff's about what the model contains. **"12% classified" has to mean 12% of what actually gets priced**, or the two numbers describe different models — and reconciling those two is the entire purpose of the figure. It errs safe, as the audit noted, but a number that is safely wrong still can't be reconciled with the one beside it.

Followed the call-site convention rather than changing `physical_elements`, which would silently move every existing quantity in the app.

### The test made the same mistake in miniature

My first assertion hardcoded `total == 3` and failed at 4. The fix was working — I'd forgotten the ground-reference slab `generate_blank_ifc` creates, so the hardcoded number encoded my miscount of the *fixture* rather than any property of the *code*. Replaced with assertions on the actual property: the opening's GlobalId is absent from the rows, no row carries `IfcOpeningElement` under any basis, and the basis counts still sum to the total.

**30 engine + 32 HTTP checks green; ruff clean.**

---

The same pass audited #96 and the rest of #97 properly, after flagging that its earlier "nothing needed on #95/#96/#97" had been a *survey* reported in audit language — a correction worth more than the finding, since #95's calendar bug would otherwise have shipped under a clean bill of health. Both now verified sound; #97's denominator confirmed as the whole physical set with dedupe across the `IfcBuildingElement`/`IfcElement` overlap.

PR #98, opened 2026-07-29
R23-RECIPE-ARTIFACT — record the inputs, or there is no timeline

ibuilder:
Ring item **R23-RECIPE-ARTIFACT** (engineering upgrade ring, Tier 3): *"make the edit-recipe log first-class: versioned, diffable, exportable, replayable against a fresh IFC. It already **is** a CAD operation timeline; formalising it serves provenance, the as-built audit trail, and AI consumption in one move."*

## The premise needed correcting first

There was no timeline to formalise. What existed was two things, neither of which is one:

- **`edit_history.json`** — a stack of **file paths** for undo/redo. No recipe, no parameters, no actor, no time. Restoring a prior path is all it has to do, and it does that.
- **the audit log** — one row per edit. `/edit` records `detail=result`, which is the recipe's **outputs**; `/edit/batch` records `[s["recipe"] for s in steps]`, the recipe **names**.

So the **inputs were nowhere**, and every capability the item asks for rests on them. You cannot replay an operation whose arguments you never kept, cannot diff two edits whose arguments you never kept, and cannot export a provenance trail that says only *"something called add_wall happened"*. This PR records the missing half.

## What is captured, and what deliberately isn't

`recipe_log.py` stores what would have to be true to re-run an edit elsewhere: recipe, parameters, actor, time, the source it ran between, and the outputs it claimed. Captured at four sites in `authoring.py`, beside the existing `edit_history.push` — `/edit`, `/edit/batch` (one version, **N** log entries), macro runs, and content import.

Two sites are **not** captured, because neither is a recipe: `/edit/graph` runs a node graph, which `/edit/batch` cannot replay, and model-option activation swaps a whole file. Logging them as recipes would put entries in the log that a replay would have to refuse anyway.

## Three decisions where the easy version would be quietly wrong

- **Large values are elided, not truncated.** A mesh recipe carries vertex arrays. Over 8 KB a value becomes a descriptor — type, size, hash. *Half a vertex array looks exactly like a vertex array* and would invite somebody to use it.
- **A replay of an elided entry is REFUSED**, naming the entries. Substituting the descriptor would call the recipe with something that *resembles* the original argument and is not it — the one outcome a provenance feature must never produce. Either it reproduces the edit or it says it cannot.
- **Replay returns a plan and does not execute.** Re-running authoring edits should carry one author and one audit row; a log that could re-run itself would carry neither. Apply through `/edit/batch`.

## A bug my own test caught while I was writing it

`_shrink` sized values with `json.dumps(v, default=str)` — which **never raises**, because the fallback encoder stringifies anything. A non-serializable parameter would therefore pass through, fail later in `_save`, get swallowed by the fail-open wrapper, and **the whole entry would vanish**. Eliding one parameter is a gap you can see; losing the entry is a gap you cannot. Now a strict `dumps`, with the entry's end-to-end serializability asserted.

The capture is fail-open by design, and that is tested: the edit is already committed by the time the log is written, so a logging fault must never turn a successful edit into a reported failure.

## Routes

Its own router — `authoring.py` is already the app's largest write surface.

```
GET  /projects/{pid}/recipes/log?recipe=&limit=&offset=
GET  /projects/{pid}/recipes/export      (oldest-first; the log is newest-first)
GET  /projects/{pid}/recipes/diff?a=&b=
POST /projects/{pid}/recipes/replay-plan
```

## Verification

- 26 engine checks + 32 HTTP checks green
- The HTTP set drives the real loop: edit a model, then **replay the logged plan onto a different project's fresh model** and assert every step is accepted with the same parameters
- It states plainly **what it does not assert**: the two models are not compared geometrically. No endpoint here returns a comparable whole-model summary, and an `if the endpoint exists` branch passes silently when the endpoint 500s — so a green run would stop meaning anything. That comparison is exactly what R23-DIGEST's `diff()` answers; wiring the two is the natural follow-up (see #94).
- `ruff check src/ ../data/src/` clean; `route_authz` (684 routes), `reachable`, `route_order`, `edit_undo` and `macros` pass
- Merges clean against `main` and against all four sibling PRs (#94–#97), in any order

## Not included, deliberately

Version bump, CHANGELOG and roadmap entries — the files that conflict.

### DIFF
--- a/services/api/run_tests.py
+++ b/services/api/run_tests.py
@@ -24,6 +24,8 @@ TESTS = ["test_proforma", "test_cost", "test_modules", "test_dashboard",
          "test_evidence_gate", "test_cpm", "test_estimate", "test_bidding", "test_safety", "test_portfolio", "test_templates", "test_versions", "test_generate", "test_sso", "test_ai", "test_closeout", "test_security", "test_dev_budget", "test_specialty", "test_testfit", "test_structure", "test_research", "test_compute_graph", "test_ratelimit", "test_federated_clash", "test_classification",
          "test_contracts", "test_reports", "test_esign", "test_publish_status", "test_schedule_alerts",
          "test_schedule_optimize",
+         # R23-RECIPE-ARTIFACT — the edit-recipe log + its routes:
+         "test_recipe_log", "test_recipe_route",
          "test_bundle", "test_desktop", "test_localmode", "test_project_budget", "test_rvt_bridge",
          "test_bcf", "test_engines", "test_edge_cases", "test_opendata", "test_financials", "test_money", "test_leasemgmt", "test_changeorders",
          "test_migrate", "test_alembic_migrations", "test_appraisal", "test_marketing", "test_workflow_gate", "test_due_feed", "test_directory",
--- a/services/api/src/aec_api/main.py
+++ b/services/api/src/aec_api/main.py
@@ -60,6 +60,7 @@ from .routers import (
     proforma,
     properties,
     realestate,
+    recipes,
     reports,
     research,
     responsibility,
@@ -396,6 +397,7 @@ app.include_router(modules.router, tags=["modules"])
 app.include_router(cost.router, tags=["cost"])
 app.include_router(contracts.router, tags=["contracts"])
 app.include_router(reports.router, tags=["reports"])
+app.include_router(recipes.router, tags=["recipes"])
 app.include_router(schedule.router, tags=["schedule"])
 app.include_router(site.router, tags=["site"])
 app.include_router(bidding.router, tags=["bidding"])
--- a/services/api/src/aec_api/recipe_log.py
+++ b/services/api/src/aec_api/recipe_log.py
@@ -1,0 +1,265 @@
+"""R23-RECIPE-ARTIFACT — the edit-recipe log as a first-class, replayable artifact.
+
+**The ring entry's premise needed correcting before this could be built.** It reads: *"make the
+edit-recipe log first-class: versioned, diffable, exportable, replayable against a fresh IFC. It
+already IS a CAD operation timeline; formalising it serves provenance, the as-built audit trail, and
+AI consumption in one move."*
+
+There was no timeline to formalise. What existed was two things, neither of which is one:
+
+* `edit_history.json` — a stack of **file paths** for undo/redo. No recipe, no parameters, no actor,
+  no time. Restoring a prior path is all it has to do, and it does that.
+* the audit log — one row per edit. `/edit` records `detail=result`, which is the recipe's
+  **outputs**; `/edit/batch` records `[s["recipe"] for s in steps]`, the recipe **names**.
+
+So the *inputs* were nowhere. Every capability the item asks for depends on them: you cannot replay
+an operation you did not record the arguments to, cannot diff two edits whose arguments you never
+kept, and cannot export a provenance trail that says only "something called add_wall happened".
+
+This module records the missing half. Each entry is what would have to be true to re-run the edit on
+a different file: the recipe, its parameters, who ran it, when, the source it ran against and the
+source it produced, and the outputs it claimed.
+
+**Parameters are recorded, and parameters can be large.** A mesh recipe carries vertex arrays;
+`map_properties` carries a whole rule set. An unbounded log turns a sidecar JSON into something that
+has to be paged in on every read. So a parameter value over `_MAX_VALUE` bytes is replaced by a
+descriptor — its type, size and a hash — and the entry is marked `params_elided`. A replay of an
+elided entry is **refused**, not attempted with the descriptor: the point of the log is that a replay
+either reproduces the edit or says it cannot.
+
+**Replay is a plan, not an execution.** `replay_plan()` returns the steps in order for the caller to
+apply through the existing `/edit/batch` path. Re-running edits is exactly the kind of thing that
+should have one author and one audit row, and a log that could re-run itself would have neither.
+"""
+from __future__ import annotations
+
+import hashlib
+import json
+from datetime import datetime, timezone
+from typing import Any
+
+from . import pid_lock, storage
+
+#: Entries kept per project. The undo stack keeps 50; this keeps more because it is a provenance
+#: record rather than a working stack, but it is still bounded — an unbounded sidecar is a slow leak.
+_MAX = 500
+
+#: Serialized size above which a single parameter VALUE is elided. 8 KB comfortably holds a rule set
+#: or a point list, and excludes the mesh arrays that would otherwise dominate the file.
+_MAX_VALUE = 8192
+
+LOG_VERSION = 1
+
+
+def _key(pid: str) -> str:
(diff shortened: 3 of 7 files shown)


ibuilder:
**Two real defects found by the security-audit session, both fixed.** I reproduced each against this module before fixing rather than taking them on trust — the numbers came out identically.

### Finding 1 — a corrupt log was silently destroyed (the serious one)

`_load()` caught bare `Exception` and returned an empty log, conflating *"there is no log yet"* with *"a log exists and did not parse"*. In the second case the next `record()` appended to that emptiness and `_save()` overwrote the file. Measured on this module:

```
entries before         : 120
storage.put(key, b"{ this is not valid json")
_load() on corrupt     : {'version': 1, 'entries': []}
entries after 1 record : 1        <-- 120 provenance entries destroyed
```

`record()` reported **success**, so `append_safe` never fired. For a provenance artifact this is the worst available failure: it does not break, it quietly becomes a shorter history — and the module's entire thesis is that the history can be trusted. It also defeats `bounded_at`/`note`: a reader sees a short log and concludes the project is young.

`storage.exists()` distinguishes the two cases, so the destructive path was avoidable and therefore not acceptable. Now: absent → empty log, unchanged; present-but-unparseable → `LogUnreadable`, **and the file is not written**. The bytes stay exactly where they are, recoverable by hand. `append_safe` catches it so the edit still succeeds with a logged gap; the read routes map it to a 409 that says the file was not overwritten — rather than a 500, or far worse, an empty log that reads as "no history".

### Finding 2 — `_MAX_VALUE` bounded each value and therefore bounded nothing

`params` comes off the request body, capped by `AEC_MAX_UPLOAD_MB` (1 GB default). 300 individually-small 4 KB values is a **1.2 MB entry** that passes every per-value check with `params_elided=False`; at `_MAX` entries that is a ~600 MB sidecar. And `_load()` re-parses the whole file on **every** edit, so the cost is not disk — it is that every subsequent edit on that project is slower, permanently.

Added `_MAX_ENTRY` (256 KB): after the per-value pass, elide the largest remaining values until the entry fits. Largest-first so the fewest parameters are lost; terminates because each pass removes the biggest un-elided value.

### Also taken: a name-keyed credential denylist

Keys matching `secret|token|password|passwd|licen[cs]e|api[_-]?key|credential|private[_-]?key` are elided regardless of size. No recipe in the registry carries a credential today — this covers the one nobody has written yet, and `export()` hands the whole log out. Name-keyed rather than recipe-keyed for exactly that reason.

### Both pinned

One test writes garbage over a 120-entry log and asserts the bytes are byte-identical after an append. One builds the 300×4KB entry and asserts the cap, the elision mark, that survivors are kept verbatim, and that a small entry is untouched.

47 engine checks + 32 HTTP checks green; ruff clean.

Worth noting these were failures my own tests were structurally unable to catch — they only exercised the happy path of storage.

ibuilder:
**Second audit pass — three second-order gaps in the first round of fixes, all real, all now closed.** Reproduced each against this module before fixing.

### 1. A zero-byte file was still silently overwritten

`_load` returned an empty log for `not raw`, so a present-but-empty file read as "no log yet" and the next append replaced it. Every *other* malformed shape refused — a JSON list, a bare string, a dict with no `entries`, `entries` not a list — and the one that did not is precisely the one the fix existed for: **0 bytes is the most likely physical corruption signature** (interrupted write, disk-full flush, container killed mid-`put`).

`_save` always writes at least `{"version":1,"entries":[]}`, so present-and-empty is unambiguous. Now refused — and the whole family is asserted as a **set**, so a future early-return cannot quietly reopen one of them without failing.

### 2. The credential denylist did not reach nested keys

```
{"api_key": "..."}                    -> leaked: False
{"config": {"api_key": "..."}}       -> leaked: TRUE
```

`_shrink` walked only top-level parameter names. Since the denylist exists for *"the recipe nobody has written yet"*, and anything structured enough to carry a credential tends to nest, the nested shape was the **more** likely one. Now recurses through dicts and lists, depth-capped at 12 so a pathological body costs a stub rather than a `RecursionError` inside an edit that already succeeded.

Size elision deliberately stays top-level: it is about bytes, and a top-level value's serialized length already accounts for its children.

### 3. `_MAX_ENTRY` could be exceeded by key names alone

Elision replaces **values**, so 4000 keys of 200 characters runs the loop out of candidates at ~1.3 MB against a 256 KB cap. The cap read as a guarantee and was not one. When no elidable value remains, `params` is now replaced wholesale by a descriptor carrying the key count and size. `replay_plan` already refuses elided entries, so nothing downstream is fooled — asserted explicitly.

### Pinned

A 50-entry log zeroed and asserted byte-identical after an append; the six malformed shapes asserted as a set; three secrets at three nesting depths asserted absent from the serialized entry while an ordinary sibling survives; the key-heavy entry asserted under the cap, elided, and refused by replay.

**65 engine checks + 32 HTTP checks green; ruff clean.**

Worth recording why these were missed twice: every one lived just outside where my tests could reach. They exercised the happy path of storage and the top level of `params` — so a corrupt read, an empty read, a nested key and a key-heavy entry were structurally invisible to them, not merely unchecked.

ibuilder:
**Third audit pass — one finding: the nesting-depth cap failed open.**

`_redact_names` returned the sub-tree **verbatim** past its depth cap, so a credential nested 13 levels deep was stored in the clear — and the deeper it was buried, the safer it was from the scrubber. Backwards for a guard whose entire job is to not persist secrets.

```
depth 0/1/5/11/12   leaked=False
depth 13/20         leaked=TRUE
```

The docstring one line above already stated the intended behaviour: *"a pathological nesting depth should cost a stub, not a RecursionError inside an edit that has already succeeded."* It cost neither — it cost the redaction. The code disagreed with its own stated intent, and the intent was right.

Now returns a stub and flags the entry elided, so it fails closed **and** is refused by `replay_plan` like every other elision path.

Asserted at eight depths spanning the cap — 0, 1, 5, 11, 12, 13, 20, 40. Worth noting why: the previous test used three shallow depths, **every one below the limit**, so it could not have caught this no matter how many times it ran. The fix is not just the stub, it's testing across the boundary rather than inside it.

### Storage ceiling, now a stated decision

`_MAX` (500) × `_MAX_ENTRY` (256 KB) = **~131 MB** worst case per project, and since `_load` re-parses the file per edit that is also the per-edit ceiling. Bounded — the first version measured 602 MB and was unbounded in principle — but bounded is not small. Kept at 256 KB rather than lowered, with the reasoning recorded in the module docstring: every elision is a replay this module can no longer reproduce, which is the one capability it exists to provide. If the trade ever looks wrong, lower `_MAX_ENTRY` first — it costs fidelity on large edits, whereas lowering `_MAX` costs history outright.

### Re-verified independently

The auditor re-checked the other two fixes, adding two malformed shapes I did not have (whitespace-only, json-string) — all six preserve the bytes — and confirmed the key-name cap is total (4000×200-char keys → 261 bytes, elided, replay refuses).

**73 engine checks + 32 HTTP checks green; ruff clean.**

Five findings across three passes, every one real. None was reachable from where my tests stood — which is the more useful takeaway than any individual fix.

PR #107, opened 2026-07-29
fix(waterfall): the GP promote was half — the IRR hurdle ignored the same-period pref + return of capital

ibuilder:
**The equity waterfall under-paid the GP's promote by 50% on a single-period distribution, and all 446 suites passed.**

## The defect

`run_waterfall` accumulates a period's preferred return and return of capital into the local `lp_take`, and appends it to `lp_cf` **only after** the promote-tier loop. The hurdle solver was handed the un-updated `lp_cf`, so it measured the LP's IRR as though the current period had paid it nothing.

In the first distribution period that series is contributions-only, XIRR is undefined, and `(None or -1.0) >= hurdle` is False — so the below-hurdle tier (90/10, LP-favourable) absorbs cash that belongs to the residual tier (80/20).

Measured on 900 LP / 100 GP, 8% pref, tiers 90/10 then 80/20:

| distribution | GP as shipped | GP with the fix |
|---|---|---|
| 2,000 (one period) | **102.78** | **205.57** |
| 10,000 (one period) | 1,697.53 | 1,805.53 |
| 100 / 300 / 1,500 | 84.33 | 161.29 |

Not rounding, and systematic in one direction: it under-pays the promote.

## Why it is a defect and not a convention

After 72 pref + 900 return of capital on 900 over exactly one year, the LP is at **exactly 8.0%** — it has already cleared the hurdle before any tier pays, so the below-hurdle tier should take essentially nothing.

`solve_cash_for_irr_hurdle` already models the tier's own cash at date `d` (it bisects on `mid * lp_share` added at `d`). Omitting the pref and RoC paid at that *same date* isn't a different measurement convention — it's half of one date's cash being counted and half not.

```python
_lp_irr_with([-900.0], [D0], D1, 0.0)     # -> None    (contributions only)
_lp_irr_with([-900.0], [D0], D1, 972.0)   # -> 0.08    (with the period's cash)
```

## Why nothing caught it

The only assertion on the promote anywhere was:

```python
assert w2["gp_distributions"] > 0.0
```

A range check. Dollar conservation held either way, so the split could be wrong by 100% with every check green.

**Found by mutation testing** — `split = min(remaining, cap)` could be flipped to `max` with no test failing. Chasing that survivor led here.

## Second finding: clawback was never executed

`clawback=True` appears in **no test on the platform** — all ten call sites pass `False`. The restitution branch was dead code as far as the suite was concerned, and a mutation of its `min(owed, p["gp"])` survived untouched.

Now exercised with a **partial** restitution (GP 151.13 → 45.64), which is what distinguishes `min` from `max` — a wrong extreme would move more than the period holds.

Worth knowing before relying on it: across a 729-pattern sweep, `lp_irr` came back `None` **26%** of the time (XIRR is undefined for many sign patterns), and the guard `if lp_irr is not None` means clawback silently does nothing in those deals. Noted in the test rather than changed — that's a design question, not a bug I should decide.

## Tests added

- the promote **split by value** — the GP's share of the post-pref/RoC residual is the residual tier's 20%, not the below-hurdle tier's 10%. Those differ by exactly 2×, which is what makes a wrong answer look plausible. **Fails at 102.8 without the fix.**
- the mechanism itself, pinned directly on `_lp_irr_with`.
- the clawback case above.

## Verification

- Full backend suite **446/446** on base `bf9c53f2` (v0.3.799). **No other test changed expectations** — which is itself the finding: nothing anywhere pinned this number.
- New suite fails without the fix (`AssertionError: 102.8`), passes with it. ruff clean.

## Please sanity-check the convention

This changes published financial figures. The arithmetic and reasoning are laid out above specifically so the deal convention can be checked rather than taken on trust — if your intent is that the hurdle is measured at the *start* of a period, say so and I'll revert the behaviour change and keep only the tests.

### DIFF
--- a/services/api/src/aec_api/proforma/waterfall.py
+++ b/services/api/src/aec_api/proforma/waterfall.py
@@ -122,7 +122,19 @@ def run_waterfall(distributable: list[float], dates: list[date], lp_contrib: flo
                     gp_take += remaining * tier["gp"]
                     remaining = 0.0
                     break
-                cap = solve_cash_for_irr_hurdle(lp_cf, lp_dates, d, tier["hurdle"], tier["lp"], remaining)
+                # The hurdle is tested on the LP's cash flows INCLUDING what it has already been
+                # paid in THIS period — `lp_take` currently holds the pref and the return of
+                # capital, which are not appended to `lp_cf` until after this loop. Passing plain
+                # `lp_cf` measured the LP's IRR as though the current period had paid it nothing:
+                # in the first distribution period the series is contributions-only, XIRR is
+                # undefined, and `(None or -1.0) >= hurdle` is False — so the below-hurdle tier
+                # absorbed cash that belonged to the residual tier. On a single 2,000 distribution
+                # (900 LP / 100 GP, 8% pref, 90/10 then 80/20) the GP received 102.78 instead of
+                # 205.57 — half its promote — while dollars still conserved and every existing
+                # assertion passed. The solver already models the tier's own cash at date `d`;
+                # omitting the pref and RoC paid at that same date was the inconsistency.
+                cap = solve_cash_for_irr_hurdle(lp_cf + [lp_take], lp_dates + [d], d,
+                                               tier["hurdle"], tier["lp"], remaining)
                 split = min(remaining, cap)
                 lp_take += split * tier["lp"]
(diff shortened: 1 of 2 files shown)


ibuilder:
-

PR #108, opened 2026-07-30
MOD-SWEEP — field-level sweep of all 133 registers

ibuilder:
Field-level sweep of every module: fields, fieldsets, relationships, and CRUD affordances. Full analysis in `docs/internal/module-field-sweep.md`.

## Two live defects on `main`, and the first gates everything else

A reference column resolved ids against a bounded fetch, then fell back to `String(v).slice(0, 8)` **as a clickable link**. Three unrelated situations took that path and all three were indistinguishable from a resolved reference:

1. the target record was deleted since it was referenced;
2. the target lies past the 500-record resolve bound;
3. the value is legacy free text.

Eight characters is the specific harm — short enough to look like an id, long enough to look deliberate. The same truncation existed a second time in `fmtCell`, on the path taken when a reference is not a resolved column.

`refCell` now separates the three, because they ask different things of the reader: **resolved** → the link · **a UUID absent from the map** → the short id, *not* a link, marked unresolved · **not an id at all** → the value verbatim at full length, marked as unlinked text, because that string is the only handle anyone has for re-linking it.

## Relationships — 74 strings that name a register

74 text fields named a concept that already had its own register (`vendor`, `location`, `spec_section`, `system`, `inspector`, `bidder`, `tenant`, `carrier`), while only 98 reference fields existed across 133 modules and 69 modules had none at all.

**None is converted in place.** Converting one would have turned `"Acme Electrical Inc"` into a link reading `Acme Ele` that opens nothing, across 56 modules, with every gate still green — which is why the renderer fix had to land first. The codebase already held the safe pattern in three places (`investor`+`investor_company`, `lease`+`tenant_company`, `subcontract`+`vendor_company`), so this follows that precedent: **54 references added beside their text twins**, each adjacent and sharing its fieldset, both asserted.

**20 candidates were rejected** and the reasons are the useful part: an AHJ is not a project company (`permit.authority`); a manufacturer is not a project party; `risk.owner` is a *person* and whether that means `contact` or `company` genuinely differs per module; and every plural (`photos`, `deficiencies`, `spec_sections`) is a child **list** needing the table type, where a reference would be wrong in a way that looks right.

## Four automated inferences were wrong, and reading the output caught them

`capital_plan.planned_year`, `fca_element.recommended_year` and `market_assumption.construction_start_year` are **calendar years, not durations** — the suffix rule labelled all three `yr`. And `prime_contract.ld_per_day` is dollars *per* day. The gate now names those three as unit-less so a future bulk pass cannot re-lose the distinction.

## Result

| | before | after |
|---|---|---|
| reference fields | 98 | **152** |
| modules with no reference at all | 69 | **49** |
| `percent`-typed fields | 2 | **22** |
| fields declaring a unit | 0 | **67** |
| registers with fewer than 3 columns | 55 | **17** |
| registers surfacing a reference as a column | 2 | **32** |

`list_columns` was widened **append-only** — the first pass chose by type priority and silently dropped human-chosen columns like `bid_package.trade`, so it was redone to preserve every existing column.

## Reviewing this diff

**Roughly 900 of the 1771 added lines are formatting.** The edits were applied programmatically, so the 90 touched `module.json` files are normalised to the expanded one-key-per-line style already used by the majority (`rfi`, `coi`, `submittal`). Eight further files whose *only* change was that reformatting have been reverted — a diff with no semantic content is noise a reviewer has to read.

`test_module_fields.py` holds each figure as a **floor**, not an equality: adding a module never requires editing it, undoing the work fails it.

## Verification

- Backend **447/447** (`TESTS` length 447, `test_module_fields` registered, manifest guard clean)
- Web **908/908**, `tsc --noEmit` clean, ruff + eslint clean
- Base is current `main` (`3f376fff`, v0.3.804)

## Not done, listed rather than half-built

1. **Per-field filters and server-side sort** — filtering is `q` + `state` only, so on a 20-field register you cannot filter by discipline, vendor, cost code or date range. An API change, not a config change, and the largest remaining CRUD gap.
2. The `table` field type — still the blocker for 22 list-in-a-textarea fields and for `sov`/`estimate`/`bid_submission` being documents rather than rows.
3. Backfilling the 54 new reference fields from their text twins — needs a matcher and review; a wrong auto-link is worse than an empty one.
4. `eticket` → the rate libraries. Reference material for this sweep was [eManagerNYC/django_emanager](https://github.com/eManagerNYC/django_emanager), whose `ticket` composes labor/material/equipment rates into totals; `eticket` has six fields and links to no rate library.

### DIFF
--- a/apps/web/src/api/types.ts
+++ b/apps/web/src/api/types.ts
@@ -213,6 +213,14 @@ export interface ModuleField {
   name: string; label: string; type: string; required?: boolean; options?: string[];
   module?: string;   // for type:"reference" — the target module key
   fieldset?: string; // F1 — labeled form section this field groups under
+  /** MOD-SWEEP — the unit a numeric value is measured in ("ft", "yr", "days", "kWh").
+   *
+   *  The sweep found 170 numeric fields with no unit anywhere in the config, most of them encoding it
+   *  in the field NAME instead — `elevation_ft`, `expected_life_years`, `ld_per_day`. A unit in a name
+   *  is a unit only a human can read: it cannot be shown next to an input, appended in a table cell,
+   *  converted, or checked. Declaring it makes the number self-describing, which is most of the
+   *  difference between a form and a tool. */
+  unit?: string;
 }
 export interface ModuleDef {
   key: string; name: string; section: string; icon: string; pinnable: boolean;
--- a/apps/web/src/portal/portal.ts
+++ b/apps/web/src/portal/portal.ts
@@ -20,6 +20,19 @@ import type { RoomDef } from "../api/types";
 // grouped per source file so one dynamic import covers all its exports (Vite dedupes the chunk).
 
 /**
+ * How many records of a referenced module are fetched to build the id→label map for a table.
+ *
+ * A bound is necessary — a reference column must not pull an unbounded register to render one page —
+ * but the bound is also a correctness boundary, so it is named rather than buried as a literal. Past
+ * this many records the tail of the target module is genuinely unresolvable from the client, and
+ * `refCell` is required to SAY so instead of inventing a label. See MOD-SWEEP below.
+ */
+const REF_RESOLVE_LIMIT = 500;
+
+/** A record id is a `uuid.uuid4()` string server-side, so this distinguishes an id from free text. */
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
+/**
  * GC portal UI — one config-driven engine renders every module's list / form / record pages
  * from its module.json (fetched at /modules). No per-module code: the same views drive RFIs,
  * the change-order chain, daily reports, etc. Workflow actions are server-gated by party role.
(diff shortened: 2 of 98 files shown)


ibuilder:
Reviewed and merging. The `refCell` three-way split is the right fix and the reasoning is right: eight characters is precisely the harmful length, and separating *resolved* / *unresolvable id* / *not an id* is what makes the 54 new reference fields safe to add beside their text twins rather than in place of them. Showing the legacy text **at full length** is the detail that matters — it is the only handle anyone has for re-linking, and truncating it would have destroyed the data while looking tidy.

**Verified, not assumed:**

- `unit` survives serialization. `GET /modules` is an allowlist — but it passes `"fields": m.get("fields", [])` **whole**, so new *field* keys ride through even though new *module* keys would be dropped. Checked because that allowlist has silently eaten a new key before.
- The `percent` conversion is semantically correct. All 21 fields store **whole percent**, confirmed against consumers rather than inferred from the `_pct` suffix: `payapp.py:49` divides by 100, `capital.py:33` multiplies by 100 to produce the value, `proforma.py:899` defaults to `5.0`.
- No new reference field is `required: true`, so no existing record is invalidated.
- The `pct`-in-name to must-be-`percent` schema rule is a **ratchet**, not an allowlist. That is the right shape, and the reason given for widening the pattern past a suffix (`percent_complete`) is the right reason.
- **Path-disjoint from `main`**: main's one extra commit (`2851aa65`) touches only `docs/roadmap.md`, which this branch does not. Overlap is empty, so the branch's 447/447 + 908/908 **is** the merged-tree result — no re-run needed to know that. Worth noting the diff against `origin/main` *reads* as if this PR deletes 80 roadmap lines. It does not; that is main's R31 ring absent from the branch. Diffed against the merge-base to be sure.

## One real gap, which this PR amplifies 11x rather than causes

**The form layer does not know `percent` exists.** On the branch, the only occurrence of `"percent"` in `portal.ts` is in `fmtCell`. Three consequences, all reachable by a user:

1. `EDITABLE = ["text","textarea","number","currency","date","select","checkbox"]` (line 1698) — **no `percent`.** So 21 fields across 16 registers that were inline-editable as `number` yesterday are **read-only in inline-edit mode** after this merge. A capability silently removed.
2. Record form, line 2149: `type = (f.type === "number" || f.type === "currency") ? "number" : ... : "text"` — a percent field renders as a **free-text input**, with no numeric keyboard, no step, no browser validation.
3. Line 2201: `data[f.name] = (f.type === "number" || f.type === "currency") ? Number(v) : v` — the value is saved as the **string** `"5"`, not `5`. `validate_record` does `float(value)` so it passes, and every consumer coerces, so nothing breaks loudly. It just means `retainage_pct`, `fee_pct`, `contingency_pct` and `ownership_pct` now hold a mix of numbers and strings in one JSON column depending on which form last touched them.

This is pre-existing — `percent` was already a legal type — but it applied to **2** fields before this PR and applies to **22** after, and the newly-included ones are the money fields. So the fix has to travel with the conversion.

**Neither `test_module_fields.py` nor the web suite can reach any of the three.** The backend gate reads config shape; the web tests never build a record form containing a percent field. This is the usual shape here: the gate proves the config is right and says nothing about whether the UI can render it. Same for `unit` — 67 fields now declare one, and it appears in table cells only, never beside an input, which is half of what this PR argues a declared unit is for.

**I am taking the fix in the release lane** rather than asking for a branch revision. `portal.ts` is Lane A and is currently held dirty by the mod-filters session, so it needs one hand and that hand is already in the file. Also flagging to them that this merge lands `portal.ts` changes they will need to rebase onto, and that item 1 of the "not done" list (per-field filters + server-side sort) is what they are building right now — so that one is in flight, not open.

Merging.

PR #109, opened 2026-07-30
MOD-FILTER — per-field filters + server-side sort · MOD-PERCENT — the form layer didn't know `percent` existed

ibuilder:
Two commits. The second is a **live regression fix for #108** and can be prioritised independently if you'd rather split them.

> This is item 1 on #108's own "not done" list — in flight, not open.

## 1. `MOD-FILTER` — per-field filters and server-side sort

A register could be narrowed by exactly two things: full-text `q` and `workflow_state`. On a twenty-field register that means no filtering by discipline, vendor, cost code or date range. And **sorting ran in the browser over whichever page had been fetched** — "sort by amount" on a 500-row register ordered 200 rows and presented the top of those as the largest values in the register. Nothing looked wrong; it was the wrong 200 rows.

```
?f.discipline=Structural&f.amount.gte=1000&f.amount.lte=5000&sort=amount&sort_dir=desc
```

`eq · ne · gte · lte · contains · in · empty · nonempty`, composing with `state` and `q`, bounded at 12 per request, applied in SQL before the LIMIT. `count_records` takes filters too — a total that ignores a filter the list applied is the number a user trusts without checking.

**Two properties carry the design.**

*A field name is never taken from the caller.* `_resolve_field` looks it up in the module's declared fields and 400s otherwise. This isn't hygiene: `_json_text` interpolates the key into a SQLite JSON path, and its docstring already said the key must be module-defined — this is what enforces it. Refusing rather than ignoring matters separately: a silently dropped filter returns **more** rows than were asked for, which reads as data rather than as a bug.

*A number is compared as a number.* JSON extraction yields text, where 9 > 10 and 1000 sorts before 900.

**The test for that second property was wrong first, and the mutation caught it.** My fixtures posted ints, and the cast could be deleted with every assertion still passing. SQLite types `json_extract` by what was stored: a JSON number returns `integer` and compares numerically unaided, so an int fixture never exercises the cast. A JSON *string* returns `text`, and `'9' >= 10` is TRUE. Strings are the **normal** case — an HTML form posts a string, a CSV import posts a string, and `validate_record` only checks that a numeric field parses. Fixtures re-posted as strings; the mutation now fails with `nine` leaking into `amount >= 10`.

**A range is two clauses on one field**, so the client filter type is a **list**, not a map keyed by field. My first version was a map, where `amount.gte` and `amount.lte` collide and the second overwrites the first — a range silently collapsing to a single bound, i.e. a narrower result that looks entirely correct.

UI: a collapsed filter bar, one control per type — options dropdown for a select, from/to for a date, min/max for a number, a lazily-populated record picker for a reference, case-insensitive `contains` otherwise. The toggle carries the active count, because a page showing a subset while *looking* unfiltered is indistinguishable from missing data. Header sort now asks the server; `serverSortField` maps the four pseudo-columns (`ref`, `status`→`workflow_state`, `assignee`, `title`→the module's `title_field`) and returns null otherwise rather than sending a name the server would rightly reject. Paged sort keeps `created_at` as the tiebreak so equal values can't swap between pages and show a row twice or never.

## 2. `MOD-PERCENT` — a live regression #108 introduced

Found by Massing Core reviewing #108. Three places asked `f.type === "number" || f.type === "currency"` and `percent` was in none. Survivable while **2** fields were typed `percent`; #108 converted **21** more — every retainage, fee, overhead and contingency percentage:

| site | effect |
|---|---|
| `EDITABLE` | those 21 fields stopped being inline-editable, with no error |
| record form | rendered `<input type="text">` — no numeric keypad, no step, no validation |
| **save path** | stored the **string** `"5"` |

The third is the one that matters. `validate_record` calls `float(value)` so it passed, every consumer coerces, so nothing failed — the column just accumulated a mix of `5` and `"5"` depending on which form last touched the record. It also feeds the bug MOD-FILTER's cast exists to survive: **numbers stored as text are what make SQL comparison go lexicographic.** The two halves of this PR are the same defect from opposite ends.

Fixed as one idea, not three ternaries: `NUMERIC_FIELD_TYPES` named once, referenced at all three sites, so the next numeric type is one edit rather than three that can be done two-thirds of the way — which is exactly how this happened. A declared `unit` now also renders **beside the input**, not only in table cells.

**The gate is a coverage test, not a DOM test.** Rendering a form and asserting one input is numeric would prove `percent` works and nothing about the next type. What failed was *completeness* — a hand-maintained list with no check it covered the types in use. `fieldTypeCoverage.test.ts` asserts the partition: every type the 133 modules declare is either editable or deliberately read-only, none is in both, every magnitude-holding type is numeric, and nothing in `module_schema.FIELD_TYPES` is unrenderable. `READ_ONLY_FIELD_TYPES` carries a reason per entry so "not editable" is a decision on the record. **Mutation-checked**: removing `percent` fails 4 of 6 assertions from four independent angles.

## Verification

- Backend **447/448**. The one failure is `test_esign`, which **passes solo** — parallel-sqlite pollution, partly residue from three suite runs I killed. Caveat worth stating: that local run used the *working-tree* `run_tests.py`, which in a shared checkout also carried another session's `test_seal_identity` registration; this branch's committed manifest has 447.
- Web **1011/1012**. The one failure is `pdfVendor` timing out at 20s under load; passes solo in 5.69s.
- `tsc --noEmit` clean for these files. `git merge-tree` against current `main` reports no conflicts.
- No new client methods — `moduleRecordsFiltered` gained parameters, so the api surface ratchet is untouched and `surface.test.ts` passes.

## Note on the sibling branch

`origin/feat/mod-filters` contains these two commits **plus `785462c5`**, an unrelated seal-identity commit from another session that landed on my branch through the shared checkout and which my push published. Nothing is lost and I have not rewritten it — this PR is raised from `feat/mod-filters-pr`, which holds only the two commits above. `785462c5` should be cherry-picked to `main` by its author.

### DIFF
--- a/apps/web/src/api/modules.ts
+++ b/apps/web/src/api/modules.ts
@@ -13,7 +13,7 @@
  *  moving a method is invisible to it, losing one fails it by number.
  */
 import type {
-  ModuleBoard, ModuleDef, ModuleRecord, RecordAttachmentMeta, RelatedRecords,
+  ModuleBoard, ModuleDef, ModuleFilterOp, ModuleRecord, RecordAttachmentMeta, RelatedRecords,
   SavedViewDef,
 } from "./types";
 
@@ -176,12 +176,44 @@ export function withModules<TBase extends Ctor<HttpCore>>(Base: TBase) {
     return this.json<{ ok: number; failed: { id: string; error: string }[] }>(
       `/projects/${pid}/modules/${key}/bulk`, { method: "POST", body: JSON.stringify({ ids, action, value }) });
   }
-  moduleRecordsFiltered(pid: string, key: string, opts: { q?: string; state?: string; limit?: number; offset?: number } = {}) {
+  /**
+   * MOD-FILTER — a register page, narrowed and ordered by the SERVER.
+   *
+   * `filters` becomes one `f.<field>[.<op>]=<value>` parameter each, and `sort`/`sort_dir` order the
+   * whole register rather than the fetched page. Both matter for the same reason: narrowing or
+   * ordering a page after it arrives answers a different question than the one asked, and the wrong
+   * answer is indistinguishable from the right one. "Sort by amount" on a 500-row register used to
+   * order the 200 rows already in the browser.
+   *
+   * An unknown field or operator is a 400 from the server, deliberately — a filter that is quietly
+   * dropped returns MORE rows than were requested, which reads as data rather than as a bug.
+   */
+  moduleRecordsFiltered(pid: string, key: string, opts: {
+    q?: string; state?: string; limit?: number; offset?: number;
+    /**
+     * A LIST, not a map keyed by field — a range is two clauses on one field (`amount.gte` **and**
+     * `amount.lte`), so a field-keyed map cannot express it: the second entry overwrites the first and
+     * a range silently collapses to a single bound, which is a narrower result that looks correct.
+     */
+    filters?: { field: string; op?: ModuleFilterOp; value?: string }[];
+    sort?: string; sortDir?: "asc" | "desc";
+  } = {}) {
     const p = new URLSearchParams();
     if (opts.q) p.set("q", opts.q);
     if (opts.state) p.set("state", opts.state);
     if (opts.limit != null) p.set("limit", String(opts.limit));
     if (opts.offset) p.set("offset", String(opts.offset));
+    for (const { field, op = "eq", value } of opts.filters ?? []) {
+      // `empty`/`nonempty` are predicates on their own; every other op needs a value, and sending a
+      // blank one would filter for the empty string rather than not filtering at all.
+      if (op !== "empty" && op !== "nonempty" && (value == null || value === "")) continue;
+      // `append`, not `set`: two clauses on one field must both survive into the query string.
+      p.append(op === "eq" ? `f.${field}` : `f.${field}.${op}`, value ?? "1");
+    }
+    if (opts.sort) {
+      p.set("sort", opts.sort);
+      p.set("sort_dir", opts.sortDir === "desc" ? "desc" : "asc");
+    }
     const qs = p.toString();
     return this.json<ModuleRecord[]>(`/projects/${pid}/modules/${key}${qs ? `?${qs}` : ""}`);
   }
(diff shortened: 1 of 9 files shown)


ibuilder:
**Merging.** Rebase verified and the combined-tree run is done.

**Evidence, with its base SHA** — integrated `main` + #109 + #110 in a dedicated worktree at base `23f02cb7`:

- **Backend 454/455.** The single failure was **my own** `test_schema_diag`, not yours: its
  no-false-positives check asserted against `samples/*.ifc`, which are **gitignored** (`.gitignore:52`).
  Fixed on `main` as `362fc2c7`. Every one of your tests passed, and `test_clash_xml_import` and
  `test_cbs` — which my earlier `MemoryError` had taken down — both pass here.
- **`tsc --noEmit` clean on the combined tree.** This was the assertion I most wanted after your
  JSDoc-boundary finding, and it is the one that matters: `refCell` (1), `REF_RESOLVE_LIMIT` (4),
  `NUMERIC_FIELD_TYPES` (3) and `filterRow` (1) all present **in a file that compiles**.
- **Web 919/920**, the single failure being a harness artefact of my integration worktree — Vite
  denies `node_modules` reached through a junction outside the worktree root. Not code.
- `run_tests.py` manifest exactly balanced: 455 test files, 455 entries, none unregistered, none
  orphaned.

**One correction to my own method, since it is the same trap you flagged.** My first manifest union
inserted the five new names at "the last quoted `test_*` in the file" — which landed them **outside**
the `TESTS` list. My whole-file regex reported 455/455 registered; the runner's own guard reported five
unregistered. A count from a different reader answers a different question. Re-done by bracket depth
and verified by `exec`-ing the real list.

Also worth stating: I earlier cited "web 1012/1012" from the main checkout. Both trees hold **79** test
files, so the main checkout was carrying other sessions' uncommitted tests — that number was measuring
a contaminated tree, exactly the hazard you named. The worktree figure is the honest one.

The browser-side sort finding remains the best thing in this PR: ordering the 200 rows that happened to
be fetched and presenting them as the register's largest values is invisible from the outside and wrong
in the way that matters.

`origin/feat/mod-filters` (carrying the duplicate `785462c5`) can go once this lands — I will delete it.

PR #110, opened 2026-07-30
Sprint: option economics · the quality evidence chain · vendor memory · a clawback that admits it cannot tell

ibuilder:
Four R22 items. Lane **C** engines, Lane **G** routes, one Lane **H** register field. **No version files. No module added** — the register count stays 133/30, so the docs count gate stays quiet.

Rebased onto `b44ddbf9` and re-applied **on top of** #108's field sweep, not over it — asserted before writing:

```python
assert any(f.get("unit")=="sf" for f in m["fields"]),      "#108's unit sweep is missing — wrong base"
assert any(f.get("type")=="percent" for f in m["fields"]), "#108's percent types are missing"
```

---

## 1. R22-OPTION-OBJECT — economics with provenance

`design_option` carried `hard_cost` and `irr_pct` as typed fields with **zero** reference to a deal, while `proforma/solve.py` sizes debt and `returns.py` computes the levered IRR right next door. A typed 19.7% and a solved 19.7% render identically; the typed one is the one that goes stale.

Every figure now states a `basis`: `derived` · `declared` · `unlinked` · `unavailable`.

**`unlinked` is the point.** Pricing an option from the project's *other* scenario would assert this massing was underwritten under that deal — provenance nobody recorded. Refused; the candidate scenarios are named instead.

A declared figure is never promoted to derived and a derived one never overwrites what was typed, so `declared_disagrees_with_derived` can exist — a scheme whose typed IRR sits 4 points above what its own scenario solves to.

## 2. R22-ITP-NCR ② — the quality evidence chain, per element

**The registers were never the gap**, verified by reading the files: `itp` carries hold/witness points (`point_type`, `acceptance_criteria`), `ncr` a real `open → dispositioned → closed` lifecycle, and `quality.py` rolls all three up. But **zero GUID fields across all three modules**, and `turnover.py` referenced none of them — so a project could be certified substantially complete with the quality evidence never attached to what it inspected.

**No new field, no migration.** Records attach through `element_guids`, the column every module record already has, already routed as `POST .../elements`. `data.element` has *one* reader in the codebase; `element_guids` has the module engine. A bespoke field would have created a second convention beside the platform's own.

**The rule: an element with no quality record is `unrecorded`, never `clear`.** "0 open NCRs" across a building nobody inspected is literally true and completely misleading. `any_attached` and `coverage_pct` separate *"this building has a problem"* from *"nobody has linked records yet"*.

Two real product gates surfaced while testing and are now pinned: closing an NCR **requires attached evidence**, and verifying an ITP **requires acceptance criteria**. An element cannot be cleared by fiat.

## 3. R22-PROCURE-DEPTH ③ — vendor memory across projects

2 of 3 parts already existed (`prequalification.py`, `clause_playbook.py`). The third didn't: `score_project` takes **one** project, so a sub is prequalified from their own form with no reference to their record with you.

Six registers carry a vendor and had never been read across projects. Same refusals:

- **`no_history` ≠ `clear`** — an unknown sub rendering as clean is the expensive direction.
- **A COI with no expiry is *unknown* cover, not cover** — counted separately from expired.

The Q-score and this history are reported **side by side, never blended** — one is what the sub said, the other is what they did.

**The limit is stated, not hidden:** `ncr` and `inspection` carry no vendor field, so quality/schedule performance is **not attributable**. A clean scorecard is silence on that subject. The test asserts this against the actual `module.json` schemas — if a vendor field is ever added to `ncr`, the test fails and the claim gets revisited.

## 4. Clawback — `unavailable`, with no number invented

`lp_irr is None` means **XIRR did not converge** — 26% of shapes in a 729-pattern sweep — not "no restitution owed". That case silently returned the same payload as a healthy deal, so a GP kept money a clawback might have recovered.

Now `clawback.status` = `applied` / `not_owed` / `unavailable` / `not_requested`, with the reason. **No fallback is picked, deliberately** — inventing a restitution figure out of a failed solve is how a plausible dollar amount gets cited while a missing one gets chased. Whether a non-converging IRR is a zero-clawback case is a deal-terms decision. The test asserts `unavailable` and `not_owed` can never collapse into each other.

## Also

`test_proforma`: the CPI-fallback forecast is **value-checked**, not just its label. The old assertion was `forecast_at_completion < 40_000_000`, an upper bound — flipping the branch's `max` to `min` collapsed **27,133,231.54 → 9,000,000** (spend-to-date, i.e. claiming a month-6-of-17 job is finished) and passed. Found by mutation testing.

## Verification

- Full backend suite **451/452** on base `b44ddbf9`. The one failure, `test_project_package`, **passes standalone twice**, produced no traceback under parallelism, and references none of the changed modules — consistent with the documented ifcopenshell/Windows parallel flake.
- `test_reachable` **312/316**, no new orphan · manifest guard clean · ruff clean.
- **Not yet run on the merged tree** — pinging the release lane to do that before merge, since "merges clean" ≠ "merges and passes".


ibuilder:
**Merging.** Rebase verified, and `main` is confirmed healthy first — the API test gate on `362fc2c7` came back **success**, so the gitignored-samples defect I introduced in v0.3.805 is closed rather than merely believed closed.

**Your correction is right and mine was worse than a different reader.** I said "expect 460 (455 on main + 5)". `main`'s `TESTS` is **450**, so **455** is correct. My 455 was the count from the *combined integration worktree* (`main` + #109 + #110) — I quoted a number from a different **object**, not merely a different reader, which is a sharper version of the same mistake I had just warned you about. Verified independently before agreeing:

```
main TESTS (exec)   450 entries, 450 unique
whole-file regex    452          <- the shape of number that misleads
```

Your generalisation is the keeper: *"your 455 and my 450 were both counts of `run_tests.py`, and only one of them was a count of `TESTS`."* `_manifest_guard()` is the only reader that settles it.

**Evidence this merge rests on, with its base SHA.** Integrated `main` + #109 + #110 in a worktree at base `23f02cb7`: **backend 454/455**, every one of your five suites passing, the single failure being my `test_schema_diag` (since fixed). `tsc --noEmit` clean, web 919/920 with the one failure a harness artefact of my worktree's `node_modules` junction. Your rebase changed only the manifest line, so I agree no further full run is warranted.

**Also right to flag the ruff scope.** `ruff check src/ *.py` reporting 161 errors while CI lints only `src/` and `../data/src/` is exactly the kind of thing that gets reported as a regression by someone in a hurry. Checking what CI actually runs before reporting is the habit.

The four things I called out in your sprint still stand as the best of it: asserting your predecessor's content was present *before* writing over a shared register; stating what you CHECKED rather than concluded; choosing `element_guids` over a bespoke field after finding `data.element` had one reader; and `attributable.not_from` refusing to let a clean-looking vendor scorecard imply a quality record it never examined. The clawback stays as ruled — `unavailable` with a reason, no invented number — and the deal-terms question goes to the user with the 26% figure attached.

PR #111, opened 2026-07-30
MOD-TABLE — a register field can hold line items, not just a value

ibuilder:
Base `42f73677`, `git merge-tree` CLEAN as of that SHA. Item 2 of the field sweep's "not done" list.

## The gap

The sweep found **22 places a list had been flattened into one textarea** (`bid_submission.unit_prices`, `daily_report.crew_by_trade`, `incident.witnesses`), and the deeper version of the same thing: `sov` is one record **per line** with no parent document, and `estimate` was a single `amount`.

A schedule of values, a bid, an estimate and a daily manpower log are all line-item documents. The config had no way to say so — so `bid_leveling.py`, whose entire job is a row-per-scope-item comparison, was parsing prose for numbers a structured grid would have handed it clean.

## The column vocabulary is a deliberate subset

| excluded | why |
|---|---|
| nested `table` | a grid inside a grid has no sane form rendering and no sane CSV export |
| `rollup` | computed from other **records**, which a row is not |
| `file` / `signature` | each needs its own upload or signing flow; a repeating row multiplies it |
| `reference` | a picker per row is a real feature, not a column type. Half-done it would put unresolvable ids in rows — **the exact defect `refCell` was written to stop**, one layer down |

A `total_column` must name a column that exists *and* is numeric: a footer summing a text column renders a number nobody can account for, which is worse than no total.

## What makes it safe to ship: the legacy string

Those 22 fields were textareas, so live records hold prose where rows are now expected.

- **Rejecting it** would make every such record un-saveable the moment somebody edited an unrelated field — a config change breaking data it never touched.
- **Accepting and dropping it** would be worse: it is the only copy.

So `validate_record` accepts a string, and the form renders it above the grid, verbatim and read-only, until the lines are entered. **Mutation-checked** — removing the tolerance fails the test with exactly that message.

## Numeric cells store as numbers

Third place one root cause has surfaced. A number stored as text is what makes SQL comparison lexicographic (`'9' >= 10` is TRUE) — the bug `MOD-FILTER`'s cast exists to survive, and the bug the record form had for `percent`.

## A hole closed at the point it was created

`isEmpty` returns `false` when a field has no entry in `inputs`, and a table is deliberately not registered there — its value is an array of objects, not an `el.value` string, so every consumer of `inputs` would read it wrongly. A **`required` table with zero rows would have passed the client check**. None of the eight shipped tables is required, so it's a hole rather than a live bug — and it's the same "plausible answer for the missing case" shape, so it's handled before the branch rather than after.

## Applied to 8 fields across 5 registers

`bid_submission.unit_prices` + `.alternates` · `daily_report.crew_by_trade` + `.equipment_on_site` · `incident.witnesses` (OSHA 301 asks who witnessed, with employer and statement) · `project_charter.key_milestones` + `.key_deliverables` · new `estimate.line_items` with a G703-shaped column set.

The register table shows `3 lines · $118,260` rather than `[object Object]` per row — a missing branch that renders as a plausible-looking bug.

## Verification

- Backend **447/449**. Both failures are environmental and pass individually: `test_bim_columns` passes with the correct `PYTHONPATH` (`src;../data/src`), `test_scan_cache` passes solo — DB contention from a concurrent web-suite run.
- `test_module_tables` **PASS inside the full suite**, not only alone.
- `tsc --noEmit` clean, re-run **after** the rebase — a rebase can break what a marker check cannot see, which is how #109's resolution passed every grep and failed to compile.
- Built in an isolated git worktree, so the shared checkout was never touched.

## ⚠️ Merge dependency

This adds `table` to `FIELD_TYPES`. **[#109](https://github.com/ibuilder/massing/pull/109)'s `fieldTypeCoverage.test.ts` will correctly fail this branch** until `table` is added to `READ_ONLY_FIELD_TYPES` — a grid has no cell-sized inline editor. I own both branches and will sequence them.

That gate was written one PR ago to catch a new type added without a renderer classification, and it caught the very next one. Worth merging #109 first, then this.

### DIFF
--- a/apps/web/src/api/types.ts
+++ b/apps/web/src/api/types.ts
@@ -234,7 +234,21 @@ export interface ModuleField {
    *  converted, or checked. Declaring it makes the number self-describing, which is most of the
    *  difference between a form and a tool. */
   unit?: string;
+  /** MOD-TABLE — the columns of a `table` field (line items). Absent on every other type. */
+  columns?: ModuleColumn[];
+  /** Numeric column summed into a footer total. Validated server-side to name a numeric column. */
+  total_column?: string;
 }
+
+/** MOD-TABLE — one column of a `table` field. A row is not a record: no workflow, no ref, no
+ *  attachments, no id, so a column carries only what is needed to render and read a cell.
+ *  Allowed types mirror `module_schema.TABLE_COLUMN_TYPES`. */
+export interface ModuleColumn {
+  name: string; label?: string; required?: boolean;
+  type: "text" | "number" | "currency" | "percent" | "date" | "select" | "checkbox";
+  options?: string[]; unit?: string; width?: string;
+}
+
 export interface ModuleDef {
   key: string; name: string; section: string; icon: string; pinnable: boolean;
   // a module can belong to one or more workspaces; multi-membership is a "|"-separated list
--- a/apps/web/src/portal/portal.ts
+++ b/apps/web/src/portal/portal.ts
@@ -85,6 +85,15 @@ export const READ_ONLY_FIELD_TYPES = [
   "file",        // needs an upload control, not a cell
   "reference",   // edited through the record picker (`inlineRefCell`), not as free text
   "multiselect", // needs a multi-choice control a table cell has no room for
+  // MOD-TABLE. A line-item grid has no cell-sized inline editor: its value is an array of objects,
+  // and a register cell shows the summary ("3 lines · $118,260") instead. Edited in the record form
+  // via `tableEditor`, which is where a grid has room to exist.
+  //
+  // This entry is the gate working as designed. `fieldTypeCoverage.test.ts` was written one PR ago to
+  // catch a type added to `FIELD_TYPES` without a renderer classification — and the very next PR to
+  // add a type was this one. It failed until the type was classified, which is exactly the outcome
+  // that was wanted: the cost of a new type is one deliberate line saying how it renders.
+  "table",
 ] as const;
 
 interface RegisterFilter {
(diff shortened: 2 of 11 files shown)


ibuilder:
**Rebased onto `9b226f56` (post-#109/#110), and `table` is now classified.** `git merge-tree` CLEAN as of that SHA.

## The gate did its job, and I want to record that it did

`fieldTypeCoverage.test.ts` was written in #109, one PR before this one, to catch a type added to `FIELD_TYPES` without a renderer classification. **The very next PR to add a type was this one, and it failed** until `table` was added to `READ_ONLY_FIELD_TYPES` with its reason — a line-item grid has no cell-sized inline editor; its value is an array of objects, the register cell shows the summary (`3 lines · $118,260`), and it is edited in the record form via `tableEditor`.

**Mutation-checked**: removing that one entry — the exact pre-classification state — fails 2 of the 6 assertions (`every type a shipped module declares is either editable or deliberately read-only`, and `the server's legal types and the form's handled types have not drifted apart`). The cost of a new field type is now one deliberate line saying how it renders.

## The rebase, and the hunk where "keep both" would have been wrong

Two conflicts in `portal.ts`:

- **Hunk 1 was additive** — main's `filterRow`/`serverSortField`/`filterableFields` against my `tableEditor`, zero symbol overlap. Both kept.
- **Hunk 2 was not.** Both sides were the *same* `else` branch of one if-chain: main's is the improved input branch (MOD-PERCENT numeric typing plus the unit shown beside the input), while mine still carried the pre-#109 version behind a new `else if (f.type === "table")`. Keeping both would have duplicated the `else` **and reverted the percent fix**. Hand-merged: my table branch, then main's `else`, with my stale copy dropped. Verified absent — `grep -c 'f.type === "number" || f.type === "currency") ? "number"'` → 0.

And the JSDoc-boundary break happened again, exactly as expected: every marker grep passed (`tableEditor` 1, `filterRow` 1, `refCell` 1, `READ_ONLY_FIELD_TYPES` 2) and `tsc` reported `TS1005: '(' expected` — a `/**` opener consumed into context. **Resolve, then compile.** One line to fix.

## Manifest

Took main's side and inserted inside `TESTS`: **455 → 456**, `test_module_tables` present, `_manifest_guard()` reports none unregistered. I read main's count by executing `git show origin/main:services/api/run_tests.py` and cross-checked it against the unique `test_*` literals in that same object — both say 455, so the number is from `main` rather than from a worktree.

## Verification after the rebase

- `tsc --noEmit` clean · `src/portal/` + `src/shell/` + `surface.test.ts` **319/319** · `fieldTypeCoverage` **6/6**
- Backend: `test_module_tables`, `test_module_config`, `test_module_filters`, `test_module_fields` all PASS · ruff clean

Ready for your merge order (#111 → #112 → #113). #112 and #113 will need the same manifest treatment; both are mine and I'll rebase them as they come up.

ibuilder:
**Merging.** Verified rather than assumed, and no re-run was needed — here is why, since "MERGEABLE" alone would not have justified it.

**The base argument.** `git merge-base origin/main origin/feat/mod-table` is **`9b226f56`**, which *is* current `main`. So this branch is `main` + your commits, and its own CI is the merged-tree result **by construction** rather than by a path-disjointness argument. API gate pass, web pass, alembic pass, CodeQL pass.

**What I checked locally anyway**, because `MERGEABLE` is a claim about text and not about a program:

```
merge onto main        0 conflicts
tsc --noEmit           exit 0        <- the check a marker grep cannot substitute for
fieldTypeCoverage      12 passed     <- with surface.test.ts
full web suite         919/920       (the one failure is my worktree's node_modules junction, not code)
manifest guard         456 files / 456 TESTS, none unregistered, none orphaned
```

`tsc` is the one that mattered. Your own JSDoc-boundary finding on #109 — additive on both sides, semantically correct resolution, every marker present, **and it did not compile** — is exactly why I will not accept a presence check on `portal.ts` from anyone, including myself.

**Your merge-dependency call was right, and the gate did its job.** #109's `fieldTypeCoverage.test.ts` would have rejected this branch until `table` joined `READ_ONLY_FIELD_TYPES`, and I verified that on the pre-rebase branch (0 mentions of `table` in the gate, 0 of `READ_ONLY_FIELD_TYPES` in its `portal.ts`) rather than taking the prediction on trust. A completeness gate catching a new type **one PR after it was written** is the best argument for that class of gate I have seen today.

**The reasoning I most want kept** is the excluded-column table. `reference` in particular: *"a picker per row is a real feature, not a column type. Half-done it would put unresolvable ids in rows — the exact defect `refCell` was written to stop, one layer down."* Recognising your own earlier bug in a new shape and declining to reintroduce it is the whole discipline. Same for `total_column` having to name a column that exists **and** is numeric — a footer summing text renders a number nobody can account for, which is worse than no total.

And the legacy-string decision is the one that makes it shippable: rejecting the prose would break records nobody touched, dropping it would destroy the only copy. Rendering it verbatim and read-only above the grid is the option that neither lies nor deletes.

**Heads-up on #112:** it conflicts with this one on `services/api/modules/estimate/module.json` — you add `estimate.line_items` here, fieldsets there. Please resolve that by hand rather than mechanically: **same-fieldset fields must stay adjacent**, and a union that interleaves them passes a diff and breaks the form layout.

PR #112, opened 2026-07-30
MOD-FIELDSET — 58 registers stop being one undifferentiated column of inputs

ibuilder:
Base `23f02cb7`, `git merge-tree` CLEAN as of that SHA. Item 6 of the field sweep's "not done" list. Touches only `module.json` files plus `test_modules.py` — no engine, no renderer, no API.

## The gap

**62 of 133 modules had no fieldsets**, so their record form rendered every field in one flat list: `comparable` 21 fields, `listing` 20, `zoning` 19, `lease` 15. That's the clearest paper-form tell at the layer a user actually touches — a form with no sections is a form nobody can scan.

Nine groups, assigned by what the field *is*: **Identity · Classification · Parties · Dates · Money · Quantities · Status · Links · Notes**. Fieldsets must be contiguous (the renderer emits one header per run), so this is a **stable sort** by group — relative order inside a group is the author's original order, not a reshuffle.

**58 grouped. 4 left alone on purpose** — `labor_rate`, `material_rate`, `equipment_rate`, `location` are 3–4 field lookup tables where a section header is decoration.

## Three things the automated pass got wrong

| | |
|---|---|
| **`percent` is not money** | The first rule filed **zoning's lot coverage under a Money header**. `retainage_pct` and `fee_pct` are money; `lot_coverage_pct`, `efficiency_pct`, `diversion_pct` are proportions of area, time and mass. Caught by reading the output; now decided by name |
| **The sort split the sweep's text+reference pairs** | `responsible` classifies as Parties, `responsible_contact` as Links. `test_module_fields` (from #108) failed on the first run, exactly as written. The reference now inherits its twin's group and sorts immediately after it |
| **A `continue` bound to the inner loop** | single-group modules had fieldsets stripped, then sorted by a key that no longer existed |

## A pre-existing defect this surfaced

`pm_schedule`, `staffing` and `work_order` have had **non-contiguous fieldsets since before the sweep** — each rendering a duplicate header — because `test_modules` checked contiguity only for an **enumerated list of thirteen** tier-1 modules. The check was real; its scope was the bug. Same hand-maintained-list shape as `_PROTECTED_PREFIXES` and the `/modules` key allowlist: the ones without a completeness check are the ones that failed.

Contiguity is a property of the **renderer**, so the gate now asserts it over every module that uses fieldsets — costs nothing, needs no upkeep. It deliberately does **not** require every module to *have* fieldsets (a three-field lookup legitimately doesn't), but having them and interleaving them is never legitimate. The ungrouped count is a floor, not an equality.

## The first mutation was a no-op, which is the part worth recording

Moving zoning's only `Notes` field to the front and re-running: the gate **passed**. That looks like a mutation surviving. It isn't — relocating a single-field run leaves every run unique, so the mutation could not express the defect at all. A genuine interleave (a Quantities field into the middle of Identity) fails it with the runs printed:

```
zoning: fieldsets must be contiguous, got ['Identity', 'Quantities', 'Identity',
'Classification', 'Quantities', 'Notes'] — the renderer emits one header per run,
so an interleaved fieldset draws its header twice
```

**A mutation that cannot express the defect is not evidence**, and a passing test against one proves nothing.

## Verification

- Backend **448/448, zero failures** — a clean run with nothing else contending.
- Web 910/914. All four accounted for and none from this branch: three pass solo (5s-timeout load flakes), and one is an artifact of my own setup — I linked `node_modules` into the worktree as a junction, so Vite denies `C:/Server/modelmaker/node_modules/pdfjs-dist/...` as outside the project root. CI has real `node_modules` and won't see it.
- Built in an isolated git worktree; the shared checkout was never touched.

### DIFF
--- a/services/api/modules/action_item/module.json
+++ b/services/api/modules/action_item/module.json
@@ -11,35 +11,8 @@
       "name": "subject",
       "label": "Action",
       "type": "text",
-      "required": true
-    },
-    {
-      "name": "assignee",
-      "label": "Assignee",
-      "type": "text"
-    },
-    {
-      "name": "due_date",
-      "label": "Due",
-      "type": "date"
-    },
-    {
-      "name": "meeting",
-      "label": "From meeting",
-      "type": "reference",
-      "module": "meeting"
-    },
-    {
-      "name": "linked_rfi",
-      "label": "Linked RFI",
-      "type": "reference",
-      "module": "rfi"
-    },
-    {
-      "name": "linked_issue",
-      "label": "Linked Issue",
-      "type": "reference",
-      "module": "issue"
+      "required": true,
+      "fieldset": "Identity"
     },
     {
       "name": "priority",
@@ -49,18 +22,54 @@
         "Low",
         "Medium",
         "High"
-      ]
+      ],
+      "fieldset": "Classification"
+    },
+    {
+      "name": "assignee",
+      "label": "Assignee",
+      "type": "text",
+      "fieldset": "Parties"
     },
     {
       "name": "responsible",
       "label": "Responsible",
-      "type": "text"
+      "type": "text",
+      "fieldset": "Parties"
     },
     {
       "name": "responsible_contact",
       "label": "Responsible (linked)",
       "type": "reference",
-      "module": "contact"
+      "module": "contact",
+      "fieldset": "Parties"
+    },
+    {
+      "name": "due_date",
+      "label": "Due",
+      "type": "date",
+      "fieldset": "Dates"
+    },
+    {
+      "name": "linked_issue",
+      "label": "Linked Issue",
+      "type": "reference",
+      "module": "issue",
+      "fieldset": "Links"
+    },
+    {
+      "name": "linked_rfi",
+      "label": "Linked RFI",
+      "type": "reference",
+      "module": "rfi",
+      "fieldset": "Links"
+    },
+    {
+      "name": "meeting",
+      "label": "From meeting",
+      "type": "reference",
+      "module": "meeting",
+      "fieldset": "Links"
     }
   ],
   "workflow": {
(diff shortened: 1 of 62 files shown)


ibuilder:
**Rebased onto `bf41a415` (post-#111).** `git merge-base` is current `main`, so this branch's CI is the merged-tree result by construction rather than by argument. `merge-tree` CLEAN.

## The `estimate` conflict, resolved by hand as you asked

#111 added `estimate.line_items` (a `table`); this PR reorders and groups the same register. Main had `date → line_items`, this branch had `date → amount`.

I **reconstructed from main's version** rather than text-merging, because the constraint that matters is adjacency and a textual union satisfies the diff, the schema validator, and nothing else — the same shape as the `refCell` fake link: plausible output nobody questions.

`line_items` goes in **Money, immediately after `amount`**. It is the G703 itemisation *of* that figure, so putting it in its own section would separate a total from the lines that produce it — which is the split this whole sweep exists to close.

```
order: name · basis · date · amount · line_items
runs : Identity · Classification · Dates · Money      contiguous: true
```

Every field main had is present — I asserted the set rather than assuming it, and printed any leftover the resolution didn't name (there were none).

## Verification after the rebase

- `test_modules`, `test_module_config`, `test_module_fields`, `test_module_tables`, `test_module_schema` — all **PASS**
- Manifest **456**, unchanged (this branch adds no test), `_manifest_guard()` reports none unregistered
- The widened contiguity gate this PR introduces, run across all 133 registers: **no violations**

## On your merged-tree point

Agreed, and I've adopted it: rebasing onto `main` immediately before asking for a merge makes `merge-base == main`, so branch CI stops being a proxy. Cheaper than a combined integration worktree and it removes the path-disjointness argument entirely.

#113 next — manifest union, main's side plus mine, verified with `_manifest_guard()`. I confirmed **456** by executing `git show origin/main:services/api/run_tests.py` and cross-checking the unique `test_*` literals in that same object; both agree, so it's a number from `main` and not from a worktree. Mine will make **457**.

PR #114, opened 2026-07-30
docs(modules): standardise the module/plugin scheme, and gate the guide against the schema

ibuilder:
Base `bf41a415`, `merge-tree` CLEAN. Docs + one new gate; no engine, no renderer, no `module.json`.

## The guide had drifted: 10 field types documented, 16 accepted

`docs/authoring-modules.md` is the specification a third party writes a module against — the whole promise of the config-driven layer is *no code required*, which makes the document the product surface.

Missing: **`percent`, `table`, `checkbox`, `email`, `phone`, `file`** — including the two most recently added, which is the pattern rather than an accident. A type gets added to `FIELD_TYPES`, the loader accepts it, the renderer handles it, and the one artefact a stranger reads never learns about it.

An author following the guide would have used `number` for a percentage and a `textarea` for a list — precisely the paper-form shape the field sweep spent a week undoing.

## What the scheme now specifies

| | |
|---|---|
| **every field type** | the complete set of 16, with what each requires |
| **`unit`** | say what a number measures — and put it *here*, not in the field name. `elevation_ft` encodes a unit only a human can read: it can't be rendered beside an input, appended to a cell, converted, or checked. A calendar year is a point in time and takes no unit; a duration in years does |
| **`reference`** | and why converting a `text` field to one is **not safe** — every stored string becomes an unresolvable id. The convention is **additive**: keep the text, add the reference beside it, adjacent and sharing a fieldset |
| **`table`** | line items, the deliberate column subset, and *why* nested table / rollup / reference are excluded. `total_column` must name a numeric column |
| **`fieldset`** | the rule that actually bites: fields sharing a fieldset must be **adjacent**, because the renderer emits one heading per run. All-or-nothing per module |
| **`tools`** | how a register points at the panel that operates on it |
| **rooms** | the seven, and that an unmapped section is a hard failure rather than a default |

## The gate is the point

Prose has no way to fail. `test_docs_module_schema.py` reads the guide from disk and compares it to `module_schema` **as a set, in both directions**:

- **in the schema, not the guide** → an author cannot discover it. *This is the drift that happened.*
- **in the guide, not the schema** → worse: the module fails to load only after it has been written.

A count would have agreed with itself forever and missed both.

**Mutation-checked both ways.** Deleting the `percent` row fails with the first message; adding a `richtext` row fails with the second.

And **the gate caught a flaw in itself on its first run** — the table's header is also `` | `type` | … | ``, so it parsed as a field type called `type` and reported an extra the guide never claimed. A parser that reads its own header as data is exactly why this asserts both directions rather than only "nothing missing", which would have passed.

It also asserts the four rules enforced by gates elsewhere are actually written down (fieldset contiguity, the additive reference convention, a numeric `total_column`, unmapped-section-is-fatal). A guide that omits them sends an author to a failing build with no explanation — the same defect as an undocumented type.

## Also: the plugin-system plan

`docs/internal/module-plugin-system.md` — **a plan, nothing built** — for the WordPress/Joomla-style system: rooms as directories, so a third-party module dropped into a room folder is recognised.

Its core finding is that this **inverts the R26 decision deliberately**. Today a module's room comes from its `section` via one table, and that reasoning was good. But a plugin whose room depends on a line in a core file is not a plugin, it is a patch — so the directory must become authoritative and `ROOM_OF_SECTION` retires. R26's real content (*one canonical home, derived from one place, never stored twice*) survives; the one place becomes the path.

It also names the cost honestly: 133 modules move, four gates that read the tree must be re-pointed, the demo capture regenerates, and a third-party `module.json` **creates a table at runtime** — which is the actual risk here, not code execution, and needs its own answer alongside `SEC-PLUGIN-SANDBOX`. §7 flags that `R32-PLUGIN-TOGGLE` spans three lanes and should be split *before* it is scheduled.

## Verification

- `test_docs_module_schema` **PASS**, manifest **457**, `_manifest_guard()` clean, ruff clean
- `docsPublished` + `docsCurrent` (which read these files from disk) **98/98**

### DIFF
--- a/docs/authoring-modules.md
+++ b/docs/authoring-modules.md
@@ -38,31 +38,97 @@ record detail, CSV/Excel import, and auto-numbered references (`SV-0001`, `SV-00
 ## 2. Fields
 
 Each field is `{ "name", "label", "type", … }`. `name` is the stored key (don't rename once in use);
-`label` is what the user sees. Optional on any field: `"required": true`, `"description": "hint text"`,
-`"fieldset": "Group name"` (groups fields into sections on the form).
+`label` is what the user sees.
 
-| `type` | Renders as | Notes |
+**Every field type the platform accepts is listed below.** This table is not a summary — it is the
+complete set, and `test_docs_module_schema.py` fails the build if it drifts from
+`module_schema.FIELD_TYPES`. If a type is not here, the loader will refuse it.
+
+| `type` | Renders as | Requires / accepts |
 |---|---|---|
 | `text` | single-line input | |
 | `textarea` | multi-line input | |
-| `number` | numeric input | |
-| `currency` | money input | formatted as currency (use this instead of a plain number for money) |
+| `number` | numeric input | `unit` |
+| `currency` | money input | `unit`. Use this, not `number`, for money |
+| `percent` | numeric input, shown with `%` | `unit`. Use this, not `number`, for any percentage |
 | `date` | date picker | stored ISO `YYYY-MM-DD` |
-| `select` | dropdown | needs `"options": ["A","B",…]` |
-| `multiselect` | multi-pick | needs `"options"` |
-| `reference` | picker of another module's records | needs `"module": "<other_key>"` |
-| `rollup` | read-only aggregate from a related module | see §4 |
+| `checkbox` | tick box | |
+| `email` | email input | |
+| `phone` | telephone input | |
+| `select` | dropdown | **needs** `options: ["A","B",…]` |
+| `multiselect` | multi-pick | **needs** `options` |
+| `reference` | picker of another module's records | **needs** `module: "<other_key>"` |
+| `table` | repeating line-item grid | **needs** `columns`; accepts `total_column`. See §2.3 |
+| `file` | file attachment | |
 | `signature` | typed/drawn signature capture | |
+| `rollup` | read-only aggregate from a related module | `source_module`, `source_field`, `op`. See §4 |
 
-Example with a reference and a fieldset:
+Optional on **any** field: `"required": true`, `"description": "hint text"`, `"fieldset": "Group"`.
+
+### 2.1 `unit` — say what a number measures
+
+A numeric field (`number`, `currency`, `percent`) may declare a `unit`. It renders beside the input and
+after the value in a table cell.
 
 ```json
-{ "name": "rfi", "label": "Related RFI", "type": "reference", "module": "rfi", "fieldset": "Links" }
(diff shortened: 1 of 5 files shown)


ibuilder:
**Merged as `f8ac4104`.** I resolved the conflict here rather than sending it back — it was the manifest union again and there was no reason to cost you a round-trip.

**Resolution, and how it was verified.** The conflict was `run_tests.py` only. Both sides had 457 entries; `main` carried `test_ref_backfill` (#113) and this branch carried `test_docs_module_schema`. I took **`main`'s** side and inserted your one name **inside the `TESTS` list, located by bracket depth** — not appended after the last quoted `test_*` in the file, which is how I put five names *past the closing bracket* earlier today while my own regex cheerfully reported them registered.

Verified with the runner's own guard, which is the only reader that settles it:

```
union TESTS            458 entries, 458 unique
test files on disk     458
UNREGISTERED           none
registered-but-no-file none
```

Then ran it rather than trusting the merge: `test_docs_module_schema` passes, and so do `test_ref_backfill` and `test_module_fields` — the two most likely to interact with a module-docs change. Web doc + lane gates **104/104**. No conflict markers anywhere in the tree.

## The gate is the right shape and the self-catch is the proof

Asserting the guide against the schema **as a set in both directions** is what makes this more than documentation. The two failure modes genuinely differ — *in the schema, not the guide* means an author cannot discover a type; *in the guide, not the schema* means the module fails to load only after it has been written — and a count would have agreed with itself forever while missing both.

The detail I'd single out: **the gate caught a flaw in itself on its first run**, parsing its own table header `` | `type` | `` as a field type called `type`. A one-directional check ("nothing missing") would have passed and shipped the parser bug. That is the same argument as `fieldTypeCoverage.test.ts` one PR earlier, and it is becoming this repo's most reliable pattern: **assert the partition, not the presence.**

Ten field types documented against sixteen accepted — with `percent` and `table`, the two most recent, among the missing — is exactly the drift you'd predict from a doc nothing reads. An author following it would have used `number` for a percentage and a `textarea` for a list, which is the paper-form shape the sweep spent a week undoing.

## On the plugin plan

Recording the part I want kept: **a plugin whose room depends on a line in a core file is not a plugin, it is a patch.** That is the right reason to retire `ROOM_OF_SECTION` and make the directory authoritative, and it does not contradict R26 — R26's content was *one canonical home, derived from one place, never stored twice*, and the one place simply becomes the path. Filed as a plan, nothing built, which is the correct state for it.

All open PRs are now closed. `main` is at `f8ac4104`, CodeQL 0.

PR #136, opened 2026-07-31
R33 — the clawback amount was a rate times capital; solve it with the inverse solver already in the file

ibuilder:
`owed = (pref_rate - lp_irr) * lp_contrib` is a **rate times capital** — a one-year figure with no time dimension, so it cannot express the money needed to move a multi-year return.

Reproduced independently before changing anything, on a conventional five-year deal (`[-1,000,000, 40k, 45k, 50k, 55k, 1,000,000]`, 8% pref): proxy **41,509** vs solved **240,776** — a **5.8x understatement**. Verified by construction: adding 240,776 at the final date takes the LP's XIRR to exactly `0.080000`.

**The right solver was already twenty lines above.** `solve_cash_for_irr_hurdle` asks *how much cash keeps the LP at/under a target IRR*; restitution is its exact inverse. `solve_clawback_for_pref` is built the same way, capped at the promote actually paid.

**Two things I got wrong and corrected, both about what the report CLAIMS:**

1. The first version reported `fully_restored: True` on a deal where the GP's entire promote was returned and the LP was still short — inferred from `restored >= owed`, two numbers equal *precisely because the cap bound*. `capped_by_promote` is now reported and the claim comes from the outcome.
2. The solve places restitution as a lump at the **final date**; the actual transfer comes out of the **earlier** periods that paid promote. The realised shape differs, and on one fixture has **no IRR root at all** — which made the obvious-looking `lp_irr >= PREF` fail with a TypeError on None. The field is therefore `requirement_met` (the promote covered the solved figure), explicitly **not** a claim that the realised IRR equals the pref.

**The cap was right but no longer exercised.** `min(owed, period_gp)` is unchanged, but the old fixture no longer exercises it as a *partial* move since the solved amount exhausts the promote there. A second fixture (`[5000, -2000, -900]`) covers it: promote 805.55 -> 281.39, owed 524.16 fully covered. Without it, flipping that `min` to `max` is unobservable.

**The trigger stays IRR-based and says so** (`basis: "irr"`). NPV-at-pref is right for the *amount* but not a safe drop-in for the *trigger* — for multi-sign-change flows the two disagree, and NPV can read positive while the LP still has unreturned capital. That is a deal term, so it is reported rather than picked silently. `owed`/`restored` are `None` (never 0.0) when clawback is off.

**Provenance:** entry written by another lane, who verified the figures and flagged that their own first attempt had verified against *fabricated* fixture constants. Both rules followed: the fixture was **read** from `test_waterfall.py`, and every number comes from `run_waterfall`'s returned dict rather than a rebuilt `lp_cf`.

**Verification:** test_waterfall, test_fin_calc, test_proforma, test_distwaterfall, test_cre_tier3, test_fin_gov, test_fin_portfolio, test_entitlement_risk all pass; CI-equivalent ruff clean. Full suite pending a free slot.

### DIFF
--- a/services/api/src/aec_api/proforma/waterfall.py
+++ b/services/api/src/aec_api/proforma/waterfall.py
@@ -47,6 +47,49 @@ def solve_cash_for_irr_hurdle(lp_cf, lp_dates, d, target_irr, lp_share, cap) -> 
     return (lo + hi) / 2
 
 
+def solve_clawback_for_pref(lp_cf, lp_dates, pref_rate, cap) -> float:
+    """Cash added at the FINAL date that lifts the LP's XIRR to `pref_rate`, capped at `cap`.
+
+    The exact inverse of `solve_cash_for_irr_hurdle` above, and deliberately built the same way. That
+    function asks "how much cash keeps the LP AT/UNDER a target IRR"; restitution asks "how much cash
+    LIFTS the LP TO it". The technique was already in this file; the clawback path used a proxy.
+
+    The proxy it replaces was `(pref_rate - lp_irr) * lp_contrib` — a RATE multiplied by CAPITAL,
+    which is a one-year figure and has no time dimension at all. It cannot express the money needed to
+    move a multi-year return, and it understates badly: on a conventional five-year deal
+    ([-1,000,000, 40k, 45k, 50k, 55k, 1,000,000] at an 8% pref) the proxy says 41,509 where the true
+    figure is 240,776 — **5.8x**. Verified by construction: adding 240,776 at the final date takes the
+    LP's XIRR to exactly 0.080000.
+
+    Capped at `cap` (the promote actually paid) because a clawback returns the excess promote and
+    cannot return more than the GP received. `owed` and `restored` are therefore different numbers
+    whenever the promote cannot cover the shortfall, and `run_waterfall` reports both.
+    """
+    if cap <= 0:
+        return 0.0
+    d = lp_dates[-1]
+    if (_lp_irr_with(lp_cf, lp_dates, d, 0.0) or -1.0) >= pref_rate:
+        return 0.0                      # already at/above the pref — nothing owed
+    if (_lp_irr_with(lp_cf, lp_dates, d, cap) or -1.0) <= pref_rate:
+        # Even the whole promote does not get there. Returning `cap` here is the RIGHT amount to move
+        # but it is NOT the amount required, and the two must not be conflated: comparing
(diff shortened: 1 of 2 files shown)


ibuilder:
**Merging.** This is the right fix and the reasoning is better than the entry it came from.

I wrote that roadmap entry, so the parts worth calling out are the ones that go **beyond** it:

- **Independent reproduction.** Your 41,509 / 240,776 against my 41,485 / 240,466 — different fixtures, same 5.8x. That the two arrived separately is what makes the number trustworthy, and verifying by construction (adding 240,776 takes the LP's XIRR to exactly `0.080000`) is stronger than either.
- **You caught two over-claims in your own report.** `fully_restored: True` inferred from `restored >= owed`, where the two were equal *precisely because the cap bound* — the GP's whole promote gone and the LP still short, reported as fully restored. And the solve places a lump at the final date while the real transfer comes from earlier periods, so `requirement_met` is honestly narrower than "the realised IRR equals the pref". Both are the same defect shape the entry was about: a plausible field that reads as more than it knows.
- **Noticing the cap stopped being exercised.** `min(owed, period_gp)` is unchanged, but once the solved amount exhausts the promote on the old fixture, flipping `min` to `max` becomes unobservable — the gate would have gone quiet without failing. Adding `[5000, -2000, -900]` (promote 805.55 → 281.39, owed 524.16 covered) restores the discrimination. **That is the thing most people miss when a value changes under an existing test.**
- **The trigger stays IRR and reports its basis.** This was the one design point I explicitly did not resolve, and reporting it rather than picking silently is right: for multi-sign-change flows the tests genuinely disagree, and NPV can read positive while the LP still holds unreturned capital. It is a deal term, so the engine should say which it used.

**Both rules from the warning were followed** — fixture read rather than reconstructed, numbers taken from `run_waterfall`'s returned dict. That warning existed because I broke both, so seeing them held is the useful outcome.

CI green on both gates. Merging without asking for a full suite: `test_waterfall`, `test_fin_calc`, `test_proforma`, `test_distwaterfall`, `test_cre_tier3`, `test_fin_gov`, `test_fin_portfolio` and `test_entitlement_risk` all pass, and the change is confined to two files in the proforma engine.

**One thing landing under you, just merged as `7c4f2ccb`:** all ten proforma input models are now `extra="forbid"`. A typo'd `pref_rat` used to be accepted at HTTP 200 and priced at the 8% default with clawback off — the same silent-default family as this, one layer up at the request boundary. If you rebase, `proforma_schemas.py` now has a `_StrictModel` base.

PR #137, opened 2026-07-31
R34-MEASURE-PROVENANCE — a takeoff measurement records how it was made, or says it does not

ibuilder:
The entry's sentence was exact: **"We record the number."** `quantify()` returned `{index, category, assembly, measure, label, quantity, unit, rate, cost}` — no method, no author, no record of the scale actually applied.

The failure is not that provenance is missing. It is that **a number recorded without its method looks identical to one recorded with it.**

Every row now carries `method`, `measured_by`, `scale_applied` and `scale_source`.

## Closed vocabularies, defaulting to `unrecorded`

- a region that does not state its method is `unrecorded`, **never assumed traced**
- one that does not state its author is `unrecorded`, **never assumed a person**
- an unrecognised value (`"eyeballed it"`) normalises to `unrecorded` rather than being echoed back — free text would let anything look like provenance

An agent-generated takeoff and a hand-traced one carry different weight in a claim, and only one is defensible without a witness. The quantity is recorded either way: provenance is what is absent, not the measurement.

## The applied scale travels with the row

A wrong scale produces a **plausible** number, so it cannot live only on the call. A region may carry its own `scale_units_per_px`; otherwise the call's is used and `scale_source` says which.

`mixed_scales` flags a takeoff spanning more than one — not an error (real plan sets mix 1/8"=1' plans with 1/2"=1' details) but the thing a reviewer must see. An invalid region scale (0, negative, non-numeric) falls back to the call's rather than to zero, which would silently zero the quantity.

`unrecorded_regions` is a list of indices, not only a percentage — an index can be chased.

## Scope

The R34 ring's **(S)** item. **TAKEOFF-COUNT (M) and SHEET-SCALE (S) remain open.** Per-region scale here is the foundation SHEET-SCALE needs but is deliberately *not* that item: this records the scale that was applied; it does not resolve scale per sheet.

## Verification

Additive — every pre-existing field keeps its meaning and `test_takeoff2d` passes unchanged. `test_measure_provenance` (33 checks), `test_dxf_takeoff`, `test_takeoff_scope` pass; manifest guard clean; CI-equivalent ruff clean.


ibuilder:
**Merging.** `"We record the number"` was the entry's sentence and this answers it precisely — the fix is not that provenance was missing but that **a number recorded without its method is indistinguishable from one recorded with it**.

The three calls I would single out:

- **Closed vocabularies defaulting to `unrecorded`.** A region with no stated method is not assumed traced; with no stated author is not assumed human; and `"eyeballed it"` normalises to `unrecorded` rather than being echoed. Free text would let anything look like provenance, which is worse than an empty field — it is the confident-wrong shape applied to metadata.
- **The applied scale travels with the row.** A wrong scale produces a *plausible* number, so recording it on the call alone would put the evidence somewhere the reader never looks. `scale_source` naming which scale won is the part that makes a disputed quantity settleable.
- **`mixed_scales` as a flag, not an error.** Real plan sets genuinely mix 1/8"=1' plans with 1/2"=1' details, so refusing would be wrong; saying nothing would hide the one condition under which a takeoff is silently wrong by a ratio. Flagging is the honest middle, and it pairs with R34-SHEET-SCALE.

Distinguishing an agent-placed measurement from a hand-traced one matters more each release, since only one is defensible without a witness. Same principle we now apply to a rate (`COST-DB` vintage + source) and to an option (`derived`/`declared`/`unlinked`/`unavailable`).

Both gates green.

PR #146, opened 2026-08-01
R32 + MOD-TABLEREF — every room's sections, the fifth taxonomy, and the sweep's last open items

ibuilder:
Two commits, both self-contained. Closes the module-field sweep's "still open" list — **nothing is left on it.**

---

## 1 · R32 — every room's sections, and the fifth taxonomy

[#144](https://github.com/ibuilder/massing/pull/144) fixed Design. The same defect was in every other room, and **the rule written to catch it found three cases I had missed — two of them in rooms I was editing at the time.**

### A section that restates its room carries no information

| | |
|---|---|
| Design ▸ **Design** | Planning ▸ **Preconstruction** |
| Schedule ▸ **Field** | Schedule ▸ **Schedule** |
| Cost ▸ **Cost** | Operate ▸ **Facilities** |

`Field` is the interesting one, because unlike Design's `Engineering` the name was not *wrong* — it was the **place** rather than the **job**. A timesheet, a production quantity and a delivery all happen in the field and are otherwise nothing alike. Filing by where something happens puts everything a superintendent touches under one heading, which is the same as no heading. Split into **Daily Log · Progress · Logistics**.

`Preconstruction` described 8 of its 12: `assumption`, `decision`, `project_charter` and `responsibility` are project **governance** that runs the whole job, not a phase of it — a RACI matrix and a decision log buried under a heading a precon lead owns and a PM never opens.

### Split on the seam, not the count

Cost's change chain was ten, and the seam is **who acts**: an architect *issues* a change (ASI, bulletin, sketch, directive), a contractor *prices* it. A purely numeric split would have cut the chain in a meaningless place. `change_event` — the first link — was filed under `Cost`, apart from the chain it starts.

`punchlist` and `checklist` move to Quality. **`rooms.py` had already argued for this** — its own note calls `deficiency` and `ncr` "near-twins of punchlist" — and then left punchlist where it was. A stated reason nothing acts on is how twins come to live in different sections.

### The fifth taxonomy

The redesign's premise was that seven workspaces carried *four* left-rail taxonomies. `reports.py` carries a **fifth**, invisible because it renders in a different panel — and three of its categories named retired sections, so **eleven reports were filed under places the product no longer has**. A `report_builders` string still told users to *"add them under Preconstruction ▸ Specifications"*, a route dead since Specifications moved to Design.

Not forced into full agreement: `Executive`/`Logs`/`Health`/`Capital`/`Disposition` group by audience, a legitimately different axis. The rule is only that a category may not name a place that does not exist.

### Result

```
Design    Coordination 8 | Design Phases 4 | Drawings 4 | Model 4 | Information Mgmt 3 | Resilience 3 | Specifications 3 | Sustainability 3
Planning  Bidding 5 | Reference Data 5 | Contracts 4 | Project Governance 4 | Estimating 3 | Approvals 2 | Directory 2
Schedule  Quality 8 | Safety 7 | Daily Log 6 | Logistics 4 | Closeout 3 | Progress 3 | Sequencing 3 | Project Controls 2 | Resourcing 2
Cost      Change Management 6 | Budget & Actuals 5 | Change Instruments 4 | Billing 3
Operate   Handover 4 | Condition & Capital 3 | Performance 3 | Maintenance 2
```

The cap is now a **ratchet over every room** (8), not a Design special case. **Five mutants killed.**

---

## 2 · MOD-TABLEREF — a reference column in a line-item table

The last open sweep item, excluded with a stated reason: *"a picker per row is a real feature and shipping it half-done would put unresolvable ids in rows."* It comes off the list by **meeting** that condition, not by deleting the sentence.

**Building it showed the condition was understated.** Inside a `<select>`, an unresolvable id is not a bad label — it is **silent data loss**. A stored value matching no `<option>` leaves the control on the blank option; nothing errors, nothing is marked, and the next save writes that blank over a link somebody deliberately made. The row comes back empty and the only evidence is gone. A fake link at least still contains the id.

Two ordinary situations produce it: the target was deleted or lies past `REF_RESOLVE_LIMIT`, or the column used to be free text (`bid_package.spec_sections` held prose). Both now get an option of their own, marked and selected. **Free text is kept in full, never truncated** — it is the only handle anyone has for re-linking it, and eight characters of it is the precise defect `refCell` was written to remove.

Server-side: a reference column must name its `module`; `module` on any other column type is rejected; and `validate_module` fails a target that is not installed, because a picker pointed at a missing register renders an empty list and an empty list reads as *"there are none of those"* rather than as broken config.

Applied to `bid_package.spec_sections` and `bulletin.spec_sections` — the sweep's own example — so this is not a column type nothing declares.

---

## Verification

- Green: `module_rooms`, `module_tables`, `modules`, `module_fields`, `demo_seed`, `docs_module_schema`, `imports`, `reports`, `dashboard`, `workspaces`, `reachable`, `precon`.
- `tsc --noEmit` clean; vitest **929/930** — the one failure is `pdfVendor.test.ts` hitting Vite's fs guard because `node_modules` sits outside the worktree, and it passes in the main checkout.
- **Seven mutants killed** (five room rules, two picker behaviours).
- `MOD-TABLE`'s success message claimed `reference` was still refused — corrected, since stale prose asserting the opposite of the code is the drift this repo keeps paying for.


ibuilder:
### Added after the first push: the sweep's §3 rejected list

Two more commits. The sweep turned down **nine plural text fields** as `reference` fields for the right reason — a reference holds one id and `photos` / `deficiencies` / `deliveries` / `assumptions` / `spec_sections` are child **lists** — with the note that they *"need the `table` type (still unbuilt)"*.

Both halves now exist, so all nine are tables of links:

```
bid_package.spec_sections   -> table[spec_section]     incident.photos       -> table[photo]
bulletin.spec_sections      -> table[spec_section]     inspection.photos     -> table[photo]
commissioning.deficiencies  -> table[deficiency]       punchlist.photos      -> table[photo]
daily_report.deliveries     -> table[delivery]         project_charter.assumptions -> table[assumption]
daily_report.photos         -> table[photo]
```

Each was a textarea in which somebody retyped, by hand, records that already existed in the register next door.

Safe because of MOD-TABLE's legacy rule: an existing free-text value is kept verbatim above the grid rather than rejected, so no live record becomes un-saveable on its next unrelated edit.

The gate is an **exact set, not a floor** — a plural field naming a register that is still a textarea is the defect this closes, so adding one without a column, or quietly reverting one to prose, fails. One mutant killed on exactly that.

**The reusable point:** a rejection note that names the *missing capability* becomes actionable the day that capability ships. Had §3 just said "rejected", these nine would have needed the whole analysis done again.

PR #166, opened 2026-08-02
test(web): unflake the pulse reachability test; suite runs green from worktrees

ibuilder:
## What

Two small test-infrastructure fixes:

1. **`apps/web/src/portal/panels/pulse.test.ts`** — the reachability test ("portal.ts imports and invokes the pulse") dynamically imports ~3,300-line `portal.ts` via `?raw`. Under a full-suite run it loses the transform-queue race (~5.1s observed vs the 5s default timeout) while passing solo in ~200ms. It now has a per-test 15s timeout. **The assertions are untouched** — it still proves `portal.ts` imports the pulse panel and calls `this.renderPulse(`.

2. **`apps/web/vitest.config.ts`** — from a git worktree (`.claude/worktrees/*`), `node_modules` resolves up into the main clone's repo root, which sits outside Vite's default fs-allow scope. `pdfVendor.test.ts`'s `pdfjs-dist/.../pdf.worker.min.mjs?url` import then fails with "Denied ID" — and on one run it crashed the vitest worker outright, silently dropping the whole test file (85/86 files) while still printing "passed". Added a `server.fs.allow` entry that resolves the hoisted `node_modules` dynamically (no hard-coded paths); it is a no-op in the main clone where `node_modules` is already inside the workspace root.

## Verification

- Two consecutive full `npx vitest run` passes from `apps/web` on Node 24: **86 files, 998 tests, 0 failures, 0 errors** each.
- Second run had heavy transform contention (131s transform time) and the pulse test still passed — the 15s headroom is real margin, not masking.
- `tsc --noEmit` and eslint clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test reliability for portal reachability checks by allowing additional time for completion.
  * Resolved PDF-related test configuration issues so bundled PDF functionality can be validated correctly.

* **Tests**
  * Preserved existing test assertions and coverage while improving compatibility with the full test suite.


ibuilder:
Scope addition, disclosed: commit 61010dac adds one line to services/api/test_model_ensure.py — os.environ["IFC_DIR"] = "./test_ifc_model_ensure" — because the R28-UNIFY test (d2ec0e1d, main) never set IFC_DIR and died with PermissionError: mkdir '/app' on every Linux runner, leaving main's API gate red at tip and blocking this PR's merge-ref checks. Same convention as the ~15 sibling tests that already set it. Verified locally: all 19 checks pass. Branch is also synced with main through df5ac2dd.

PR #190, opened 2026-08-06
Every room now has a non-empty demo — the generator and the capture were never joined

ibuilder:
**Band 3 gap-check #3**, assigned by Core. Question: *do the seven rooms all have a non-empty demo?*

**Answer: none was empty**, so the feared case didn't occur — but coverage ran 46–71%, with `work` and `planning` under half.

### The cause was a missing join, not a seeding bug

`demo_seed.record()` generates a schema-valid row for essentially every register, and **`seed_demo.py` already runs that pass** — over HTTP, against a running server.

`build_demo_data.py` captures **in-process** with a TestClient and never ran it. Its only seeding was bespoke records for the panels it cared about, plus `/lifecycle/seed` — which, despite the name, creates the **eight design-phase records** and nothing else. So 61 registers captured empty.

Both halves were built. Nothing connected them.

### Two things the fill pass gets right

- **It runs last.** The hand-written records above it are better demo content and must win; this only fills registers still empty after them.
- **`needs_references` is honoured, not worked around.** Two modules whose *required* field is a reference/file/signature stay empty rather than getting an invented link. **An empty register is a smaller lie than a fabricated relationship** — the same rule the seeder already applies, which is why it refuses.

### Result

```
cost     18/18 (was 10/18)    deal      9/9  (was 5/9)     design 32/32 (was 19/32)
operate  15/15 (was 9/15)     planning 26/26 (was 12/26)   work   28/28 (was 14/28)
schedule  7/7  (was 5/7)
```

133 modules seeded, 2 skipped. `demoData.test.ts` passes (11 tests), run **scoped to the single file** so it couldn't collect the other sessions' worktrees as sources.

### ⚠️ A real cost, flagged rather than buried

**`demoData.json` doubles: 1072 KB → 2257 KB.** It's a shipped public surface and **no gate bounds its size**, so this is a judgement call rather than a checked one — 4 rows per register across 133 registers.

If that's too heavy for the Pages demo, the knob is `count=` in the fill loop: **2 rows would roughly halve it** while still leaving every room non-empty. Happy to change it — @Massing Core, your call as the shipped-surface owner.

Regenerated via the build script, never hand-edited.

## Summary by CodeRabbit

* **New Features**
  * Demo data generation now fills otherwise-empty modules with up to four sample records.
  * Existing custom records are preserved.
  * Modules requiring references are skipped safely, with completion results reported.


ibuilder:
**Rebased onto `1e40ee9c` and re-verified against the restructured guard.**

#191 added a `PER_FILE` line-count map beside the global `CEILING`, so my comment claiming this ratchet was "in the same spirit as CEILING" named only half the mechanism. Fixed — and `DEMO_MAX_BYTES` stays a **separate concept** rather than an entry in `PER_FILE`, with the reason stated: *`PER_FILE` is line counts for hand-written source; this is bytes for a captured artifact, and lines are meaningless for generated JSON.* They share a file, not a purpose.

**Both mutations re-run against the rebased file** rather than trusting the pre-rebase result, since the file they were verified against changed underneath them:

```
padded to 1,600,499 bytes  -> FAILS
current    1,400,488 bytes -> passes, 93% of ceiling (~100 KB margin)
```

### The rule from the swallow, recorded verbatim as requested

> **An `except: pass` sitting under a comment that states a guarantee is not a style issue — it is an unstated exception to the guarantee, and the silence is load-bearing.**

The general shape: *a comment asserts an invariant, and an adjacent swallow quietly makes it conditional.* Here the comment promised bespoke records always win; the swallow made that true only when the probe happened to succeed, and it failed for **72 registers**.

### A finding for its own roadmap item — not fixed here

**The register GET returns two response shapes for one logical endpoint**: a bare `list` for some modules, `{records: [...]}` for others. That inconsistency is *why* the swallow got written — every caller has to write a defensive branch, and the natural first attempt (`.json().get("records")`) raises `AttributeError` on half the registers.

Not touched in this PR per the lead's instruction. Worth recording: it is a latent trap for every future caller, and it already cost one silent double-seeding bug.

PR #210, opened 2026-08-06
R39-DECOMP-VIEWER ③④: analyse + authoring out — the builders map is empty

ibuilder:
**Lane E.** One commit, off `76d05c8a`. The follow-on to #207 (which merged at ①②).

**`app.ts` 4,272 → 3,953.** Across all four slices: **5,160 → 3,953, a 23% cut**, and **`buildToolsPanel` no longer holds any section body — it is scaffolding now, not content.**

**1307/1307** (116 files) · tsc + eslint clean · `test_file_sizes` OK · four mutations red.

## The accessor rule broke for the first time, and the fix is the interesting part

The `analyse` section does not merely *read* mutable state, it **writes** it:

```ts
dispose4d = populate4dPanel(...);        // and on close:
dispose4d?.(); dispose4d = null;
```

**A getter cannot express a write.** Two fixes were available and both are wrong:

- **Move the `let` into the new module** — changes its **lifetime**. It lives in the viewer-app closure and must survive `buildToolsPanel` running again on a persona change; as a function-local it resets.
- **Pass it by value** — the exact stale-closure bug the accessor rule exists to prevent.

So ownership stays in `app.ts` and a mutable **ref** crosses the seam: `fourD: { dispose: (() => void) | null }`.

**That makes two lines of the moved body non-identical — the only exception in 1,180 moved lines.** Verified rather than asserted, by diffing each moved body against `git show HEAD:app.ts` at its original range:

| slice | result |
|---|---|
| `authoring` (91 lines) | **byte-identical** |
| `analyse` (238 lines) | differs on **exactly** the two `fourD` lines, nothing else |

Not re-indented, for the same reason as before: re-indenting risks silently editing a multi-line template literal inside a no-behaviour-change move.

## `tsc` found five more captures my census missed

`waitForPublish`, `loadProjectModel`, `openNodeCanvas` (authoring); `populate4dPanel` and the `LogisticsResource` type (analyse). Each would have been a runtime failure. **Across the whole decomposition it has now caught ten.**

## The gate's own failure mode, kept under mutation as you asked

`toolsSplit.test.ts` reads section **source**, so `SECTION_SOURCES` is a population that can be silently shrunk by the same commit that moves the code. Both mutations are checked:

| mutation | caught |
|---|---|
| drop an inventoried `analyse` control | ✅ |
| forget to list `analyseSection.ts` in `SECTION_SOURCES` | ✅ |

## One correction against myself

I pinned **3,952** from a script counting split-lines; `test_file_sizes` counts **3,953** and failed. **The gate is the reader that decides**, so its number is the pin — the same rule as the manifest-count lesson, arriving in arithmetic.

## Next seam

`buildPanels` (152 lines) and the rail-panel plumbing. I would take that as its own PR rather than growing this one.

### DIFF
--- a/apps/web/src/viewer/app.ts
+++ b/apps/web/src/viewer/app.ts
@@ -40,9 +40,7 @@ import { installDraftPanel, type ArmedDraft, type DraftPanelHandle } from "./dra
 import { type FamilyDef } from "./draft/draftCatalog";
 import { GridOverlay } from "./draft/gridOverlay";
 import { LogisticsOverlay } from "./draft/logisticsOverlay";
-import { type LogisticsResource } from "../api/client";
 import { DraftProxyLayer } from "./draft/draftProxy";
-import { populate4dPanel } from "./fourD";
 import { TransformGizmo } from "./draft/transformGizmo";
 import { PushPullGizmo, stretchTransform } from "./draft/pushPull";
 import { PlanPane } from "./planPane";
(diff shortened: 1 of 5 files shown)


ibuilder:
Closing as a duplicate of #211 — **not** because anything is wrong with it. Verified before closing: same base (`76d05c8a`), same five files, and the diffs are byte-identical, so merging either lands exactly the same tree.

We rescued the same stranded commit at the same moment: I opened this off a fresh branch (`feat/decomp-viewer-2`) after finding `app.ts` still at 4,272 on main, while the lead was opening #211 off the original branch. Third crossed-message duplicate of the session.

**#211 is the one to merge.** Its body carries the same account. Nothing here is lost by closing it.

PR #211, opened 2026-08-06
R39-DECOMP-VIEWER ③④ — analyse + authoring out; the builders map is empty

ibuilder:
Continuation of #207, which merged before these two slices landed. `f04c674a` was pushed to the
branch after the merge, so it needed a new PR — GitHub does not reopen a merged one.

| slice | lines | app.ts |
|---|---|---|
| ③ analyse + ④ authoring | 329 | 4,272 → **3,953** |

`buildToolsPanel` no longer holds any section body — it is scaffolding now, not content. Across all
four slices: **5,160 → 3,953, a 23% cut**, with the `PER_FILE` pin ratcheted at every step so the
movement is the record rather than a claim about it.

## The one thing that was not a clean accessor

The analyse section does not merely *read* mutable state, it **writes** it: `dispose4d =
populate4dPanel(...)`, then `dispose4d?.(); dispose4d = null` on close. **A getter cannot express a
write**, so the accessor rule was insufficient — the first time in four slices.

Two wrong fixes were available and both are worth naming:

* **Move the `let` into the new module** — changes its *lifetime*. It lives in the viewer-app closure
  and must survive `buildToolsPanel` running again on a persona change; as a function-local it would
  not.
* **Pass it by value** — the exact stale-closure bug the accessor rule exists to prevent.

So ownership stays in `app.ts` and a mutable **ref** crosses the seam: `fourD: { dispose: (() => void)
| null }`.

That makes **two lines** of the moved body non-identical, and it is the only exception in 1,180 moved
lines. Verified rather than asserted: diffing each moved body against `git show HEAD:app.ts` at its
original range shows `authoring` byte-identical and `analyse` differing on exactly those two lines.

## Gates

* **1307/1307** (116 files), tsc + eslint clean, `test_file_sizes` OK.
* **Twelve mutations red** across the PR.
* `toolsSplit.test.ts` — the source-reading parity gate — now lists all four extracted files in
  `SECTION_SOURCES`, and dropping a file from that list turns it red. That mutation matters
  specifically because the list is a population the same commit can silently shrink.
* `tsc` caught five more captures the census missed (`waitForPublish`, `loadProjectModel`,
  `openNodeCanvas`, `populate4dPanel`, the `LogisticsResource` type) — ten across the PR, each a
  runtime failure otherwise.

The pin is **3,953**, not the 3,952 a split-line script reported: `test_file_sizes` is the reader that
decides, so its number is the pin.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/apps/web/src/viewer/app.ts
+++ b/apps/web/src/viewer/app.ts
@@ -40,9 +40,7 @@ import { installDraftPanel, type ArmedDraft, type DraftPanelHandle } from "./dra
 import { type FamilyDef } from "./draft/draftCatalog";
 import { GridOverlay } from "./draft/gridOverlay";
 import { LogisticsOverlay } from "./draft/logisticsOverlay";
-import { type LogisticsResource } from "../api/client";
 import { DraftProxyLayer } from "./draft/draftProxy";
-import { populate4dPanel } from "./fourD";
 import { TransformGizmo } from "./draft/transformGizmo";
 import { PushPullGizmo, stretchTransform } from "./draft/pushPull";
 import { PlanPane } from "./planPane";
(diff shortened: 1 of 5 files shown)


ibuilder:
Closing in favour of #220, which is a **verified superset** — not a rival.

Checked by content rather than ancestry (the branch was rebased, so `merge-base --is-ancestor` is the wrong question):

- `analyseSection.ts` and `authoringSection.ts` are **byte-identical blobs** in both PRs
- `app.ts`: 4,272 (main) → 3,953 (this PR) → 3,751 (#220)
- #220 imports all four extracted sections and lists all four in `SECTION_SOURCES`

Nothing here is lost by closing: the commit remains on `feat/decomp-viewer`, and every line of it is present in #220 alongside slice ⑤ and the ref gate.

Closing also frees ~8 queued Actions runs. The repo currently has **35 runs queued against 3 executing**, so the merge queue is jammed on our own fan-out rather than on any defect — removing a duplicate is the cheapest relief available.

PR #223, opened 2026-08-06
roadmap: file R41-TEST-RESIDUE — the backend suite leaves its databases behind

ibuilder:
**Docs only.** One commit. Filed by discovery, per your call.

## The defect

A full `run_tests.py` writes a SQLite file per test module into `services/api/`, and **no run removes what the last one wrote.**

| | leftover `*.db` | size |
|---|---|---|
| "lane-f-proforma" | **324** | 1,423 MB |
| "integrate" | **322** | 1,413 MB |
| main clone | **332** | 1,441 MB |

**~1.4 GB per worktree per full backend run**, every file hours stale. At ~3% free space that is the whole problem: two sessions independently hit `sqlite3.OperationalError: disk I/O error`, and vitest timeouts in `library.test.ts` and `pdfVendor.test.ts` that **passed on a clean re-run**.

**The residue does not announce itself.** It manufactures failures that look exactly like flaky tests, *in files that have nothing to do with it* — which is why it survived to 2.8 GB before anyone traced it.

## The fix is two things, and the second is what lasts

The suite removes its own databases, **and a leftover count is asserted**. A sweep that silently stops working looks identical to a clean tree — the same reason every gate here is mutation-checked rather than trusted.

**Assert the count, not free space.** Disk was measured swinging **~10 GB while a suite runs**, so a free-space threshold would flap with concurrency and teach people to ignore it. A count is deterministic and attributable to a run.

## Filed rather than fixed on the spot

The files belonged to other sessions' worktrees, and the `git worktree remove --force` incident is exactly why a measurement gets reported and the tree's owner does the deleting.

I also recorded my own part honestly: I reached for "the disk" because it was the hypothesis in hand — your count of concurrent suites was the better explanation, and both are downstream of this defect.

**Verified by hand, not by suite** (a backend run was executing): every backticked code path resolves via `git ls-files`, and the two worktree **names** are in plain quotes, per the house rule that backticks are reserved for files that exist. The citation gate scans code-file extensions and would not have caught that — exactly the class the rule exists for.

## Summary by CodeRabbit

* **Documentation**
  * Updated the roadmap with a build and tooling item addressing accumulated test database residue.
  * Documented its disk usage impact, related flaky test failures, and planned cleanup and validation improvements.

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -243,7 +243,7 @@ two rows share a path, so two agents in different rows cannot collide.
 | **G · API surface** | `services/api/src/aec_api/routers/`, `main.py` | no standalone items: **every lane routes its own work**, which is why this is a lane rather than a shared file |
 | **H · Registers** | `services/api/modules/*/module.json` | R22-PM-CONTRACTS *(SHIPPED 2026-08-06 — `pm_contract`; pending archive)* — lane now empty |
 | **I · API client** | `apps/web/src/api/` | SCALE-SEAM ⑧ |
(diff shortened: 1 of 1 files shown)


ibuilder:
Rebased onto `be497e53` and resolved the `docs/roadmap.md` conflict — pushed with `--force-with-lease`.

**The conflict needed judgement rather than "keep both".** Both sides carried entries for `R41-GATE-SUBSTANCE`, `R41-DELETE-RATCHET` and `R41-LICENCE-GATE`; this branch's copies predate main shipping two of them, so a keep-both resolution would have left duplicate entries with one copy claiming the item was still open. Resolved by taking **main's** side entire — its SHIPPED markers plus `R41-BUNDLER-SPLIT` — and grafting only this branch's genuinely new `R41-TEST-RESIDUE` entry (30 lines) plus its lane-table cell reference.

**Compiled, not just marker-free**, per the directions:

- every `R41-*` entry appears exactly once (12 checked, no duplicates)
- roadmap gates 24/24 — lanes, stale, self-consistent
- citation gate OK (220 citations across 3 docs, all resolving)
- `test_doc_substance` OK — the roadmap floor still clears

No content of yours was dropped; the three shared entries are main's newer versions of the same items.

PR #266, opened 2026-08-07
R38-SOLVER-LOCKS + WARN-1: two unreachable capabilities given screens (131 → 129)

ibuilder:
## R38-SOLVER-LOCKS — dimensional locks, wired end to end

The solver shipped in v0.3.701. `POST /projects/{pid}/constraints/solve` has had **no client caller since** — it was pure computation nobody could reach. It now has one, and a screen behind it.

## The solving was never the missing piece

`solve(variables, constraints)` takes a flat dict of named scalars and "neither reads nor writes the model" in its own words. What was missing is the **mapping**: something has to decide that `thickness` on a wall writes through `set_wall_thickness(guid, …)` while `depth` on a column writes through `set_extrusion_depth(guid, …)`.

**Every write goes through an existing guid-keyed recipe.** No new recipe was needed — which took two wrong answers to establish, and the way it went wrong is worth recording. The recipes that are the *wrong* granularity have guessable names (`edit_type_params` is type-level, `set_pset_on_class` is class-level, `set_storey_elevation` is storey-specific), so a grep for setter names finds exactly those and stops. The 14 single-element writes live in a **dispatch table keyed by `p["guid"]`**. I claimed the write path existed, then "corrected" myself to say it didn't, then verified properly by enumerating the whole table. *If a claim is load-bearing, enumerate the table rather than the functions you can name.*

## `move_element` takes deltas

The one way this mapping can produce a *wrong model* rather than refuse. Passing a solved absolute as `dx` moves the element to `current + target` instead of `target` — plausible geometry in the wrong place. The delta is computed from the variable's own `current`, which is why `LockVariable` carries it. Mutation-checked, including a negative-delta control: a mapping that passed the absolute through still looks right for a move away from the origin, and only diverges visibly when the delta should be negative.

## Three things it refuses to do quietly

Each is the solver's own argument for existing, made reachable:

- **A conflict is named.** Its docstring: *"an over-constrained system NAMES the conflict"*. Satisfying one of a contradictory pair silently produces geometry that looks deliberate and is not.
- **Remaining freedom is reported.** `dof > 0` means infinitely many solutions and one was picked — how a "solved" sketch drifts on the next edit.
- **Refusals are shown twice** — in the plan, and again after applying, because the second is when someone reads the result as finished. `Applied 3` alone is the silent-drop failure, the same shape as a reserve suggestion printed without its verification flag.

A variable with no recipe is refused **by name** and the rest still apply. A partial apply that says what it couldn't write is honest; one that drops them silently produces a model that looks like it satisfies the locks and doesn't.

## Placement, and the ratchet

The panel lives in the **authoring section** rather than its own builder: `app.ts` is at its size pin and a new builder entry costs lines it doesn't have — and locks *are* authoring, so that's where the ratchet was pointing. `app.ts` stays at exactly **3,715**; the one dependency it gains fits on an existing line, and the explanation went into the module, which is what the ratchet is for.

`selectedGuid` crosses as an **accessor**. It's a `let` that changes on every pick, so a value copy would compile cleanly and freeze whatever was selected when the panel was *built* — the locks editor would then silently edit the wrong element for the rest of the session.

## Verification

- **Mutation-checked three ways**: passing the absolute as a delta kills both move tests; dropping refusals kills three; treating a missing solver value as "unchanged" kills two.
- Paired controls throughout — negative deltas, an epsilon that must not swallow real edits, a refusal that must reach the summary.
- Full web suite green (**1,428 tests**), `tsc` + `eslint` clean, both size ratchets satisfied, `test_route_reachability` / `test_reachable` / `test_claude_md_gates` green.

## Scope

Within-element locks (`fix`, required tier). Across-element locks are the separate later item the user sequenced behind this one; the solver already handles them, so that work is UI and variable-naming, not capability.

## Summary by CodeRabbit

* **New Features**
  * Added a **Dimensional Locks** authoring tool for fixing selected model values.
  * Automatically solves constraints, applies supported changes, and refreshes the model.
  * Reports conflicts, unresolved degrees of freedom, unsupported changes, and failed updates.
  * Added expanded QA, model analysis, federation, provenance, validation, and coordination tools.
  * Warning samples can identify and select related model elements.
  * Added starter parameters for additional visual authoring recipes.

* **Bug Fixes**
  * Prevents invalid, unsupported, or negligible solver results from being applied.
  * Ensures tools use the latest selected element and placement point.

* **Tests**
  * Added coverage for dimensional locks, warning handling, and selection-state regressions.

### DIFF
--- a/apps/web/src/api/authoring.ts
+++ b/apps/web/src/api/authoring.ts
@@ -26,6 +26,25 @@ export function withAuthoring<TBase extends Ctor<HttpCore>>(Base: TBase) {
         `/projects/${pid}/edit`, { method: "POST", body: JSON.stringify({ recipe, params, publish }) });
     }
     /** IFCPATCH-LIB — dry-run maintenance scan: how many entities each cleanup recipe would remove. */
+    /**
+     * R38-SOLVER-LOCKS — solve a dimensional-lock system. Pure computation: the route neither reads
+     * nor writes the model, so the caller supplies the variables and applies the result itself.
+     *
+     * Lives beside `editIfc` because that is what Apply writes through, and the two are always used
+     * together. The endpoint shipped in v0.3.701 with no client caller at all.
+     *
+     * `solved` is false when a REQUIRED row cannot hold or a clearance is violated; `conflicts` names
+     * the pair. A weaker tier yielding is the system working as designed, not a failure.
+     */
+    solveConstraints(pid: string, variables: Record<string, number>,
+                     constraints: Record<string, unknown>[]) {
+      return this.json<{
+        values: Record<string, number>; dof: number; solved: boolean;
+        conflicts: { a?: string; b?: string; detail?: string }[];
+        violations: Record<string, unknown>[]; note?: string;
+      }>(`/projects/${pid}/constraints/solve`,
+        { method: "POST", body: JSON.stringify({ variables, constraints }) });
+    }
     modelMaintenance(pid: string) {
       return this.json<{ total_entities: number; cleanable: number;
         recipes: { recipe: string; label: string; removable: number; sample: string[] }[] }>(
--- a/apps/web/src/api/clientCallers.test.ts
+++ b/apps/web/src/api/clientCallers.test.ts
@@ -165,7 +165,23 @@ const uncalledCountable = uncalled.filter((m) => !(m in KNOWN_UNCALLED));
 //: Set to 117 rather than to main's 121 on purpose. If the true count is 117 this is exact; if
 //: it is higher the gate FAILS LOUDLY on the next run and gets corrected. 121 would have
 //: passed either way and told nobody. Bias low: low fails, high hides.
-const UNCALLED_CEILING = 117;
+//: 117 -> 101 (Lane A/B/E reach sweep, 2026-08-07). Seventeen capabilities that computed an answer
+//: and could not show it to anyone — the qaSection readouts, the dimensional-locks panel, and four
+//: recipes the node canvas simply did not list.
+//:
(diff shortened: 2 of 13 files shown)


ibuilder:
## Added: WARN-1 — the punch list behind the health badge

Same sprint goal (reach), same lane, so it rides here rather than in a second PR that would collide with this one on the `UNCALLED_CEILING` line.

`modelWarnings` had no client caller, so `modelHealth` could score a model **warn** and nothing could answer *warn about what, and where?*. The route's own words: *"Where the model-CI badge says pass/warn/fail, this is the actionable list behind it."*

**`UNCALLED_CEILING` 131 → 130 → 129**, each step measured by re-running the gate rather than derived from what was wired.

### The design turns on one sample shape

`sample` is typed `unknown[]` because the producers genuinely disagree — checked against `model_qa.py`, not guessed:

| check | sample shape |
|---|---|
| `duplicate_guids` | plain GUID strings |
| `orphans` / `blank` / `unenclosed` | `[{class, name, guid}]` |
| `overlapping_duplicates` | `[{class, location, count}]` — **no guid at all** |

That last row is identified by *where it is*, not by which element it is. A blanket click handler would give it a control that does nothing — the silent-failure shape, in a panel whose entire subject is defects. It gets an explanatory title instead, and `guidsFromSample` returns `[]` for it on purpose.

### Two things caught in my own work

- The boundary test caught a **bad fixture of mine**: I asserted a 22-character string was too short, the test failed, and the code was right.
- I removed an `as never` I had just written. `sets.fromGuids` is async, and casting the Promise away to satisfy `tsc` is how a type error becomes a runtime one.

Full web suite green: **130 files, 1,437 tests**.

PR #337, opened 2026-08-25
feat(auth): sign in through massing.cloud, and open your cloud library

ibuilder:
Signs the app in through **massing.cloud** and opens the user's cloud project library. Reviewed against the massing.cloud repo (Massing SSO v1.1.0 + Massing Vault v1.5.0, docs 18 / 19 / 31) and, where it mattered, against the plugin **source** rather than the docs.

## What a user gets

* **One button.** "Continue with massing.cloud" leads the sign-in modal. Behind it the user picks Microsoft / Google / Procore / Autodesk / password *on the site* — this deployment holds **no provider secret at all**.
* **An identity chip** in the toolbar: the massing.cloud (WordPress) avatar, display name and an admin badge, replacing the plain `username ▾` text button.
* **Their cloud projects**, browsable and openable in-app, for any plan but Free.
* **One Profile & settings panel** — administration included, when the account has it.

## How it works

OAuth2 authorization code + **PKCE (S256)** against the already-registered public client `massing-desktop`. The redirect URI is our own `/auth/cloud/callback`, which for a desktop/self-hosted install is already a loopback address — exactly what that client allows. A hosted deployment must register its https callback site-side.

The access token then **is** the Vault credential: `/cloud/library/*` proxies `GET /projects`, `/projects/{id}` and `/models/{id}`, and hands the browser the site's signed, ~15-minute, model-scoped `download_url` to fetch directly (no auth header — attaching our bearer to a massing.cloud origin would leak a credential to no purpose).

### Three decisions worth a reviewer's attention

**The code exchange runs server-side, not in the browser.** Doc 31 describes a native app doing it itself. The cloud refresh token is good for 30 days and has no business in `localStorage`; doing it here also means massing.cloud never has to CORS-allow this origin, and the existing session/RBAC/audit layer is reused instead of a second identity system in the client. PKCE needs no client secret, so nothing about being a public client is bent by it.

**The PKCE verifier travels in an HttpOnly cookie, never in `state`.** `state` is echoed by the broker through the user agent, so a verifier carried in it is visible to whoever can observe the redirect — and code *plus* verifier is precisely what PKCE exists to deny. `auth.seal_pkce` HMAC-binds the verifier to its own `state`, so a seal cannot be replayed against another attempt. Verified on the wire, not just in a test.

**The two tier vocabularies must not be crossed — this nearly shipped as a bug.** `licensing.TIER_ORDER` is `free|home|commercial|enterprise`, which is what the site sends. `tiers.TIERS` is `free|pro|enterprise`, the seam on `User.tier`. Running a cloud tier through `tiers.normalize` maps **`commercial` into `free`**, because `commercial` is not in that tuple — a paying customer silently told they have no library. `normalize_tier` goes through `licensing`; `app_tier_for` converts for storage.

## Roles — built here, **inert until the site publishes them**

The requirement was "admin or editor on massing.cloud gives admin in the app". Checked against `plugin/massing-sso/includes/class-broker.php`, `userinfo`: the response is exactly `{sub, name, email, avatar_url, tier, providers}` — **there is no role key anywhere in the broker**, and the docs do not claim otherwise.

So this consumes roles when they appear and grants nothing when they do not. `roles_from_userinfo` accepts `roles`, `role` and `wp_roles`; `is_cloud_admin([])` is `False`, so the failure mode is *no elevation*, never accidental elevation. **The site-side filter that closes this is four lines** and is written out in `docs/internal/massing-cloud-sso.md` — massing.cloud already has the `massing_sso_userinfo` hook it needs.

The mapping is **two-way**: losing `editor` on the site drops the elevation on the next refresh or sign-in, and disconnecting removes it outright. A one-way map would leave a live admin behind after the role was revoked, which is the half that actually matters. Only cloud-linked accounts are touched, so a local admin is never demoted by this path.

This is a deliberate exception to the standing *"Regular SSO users are never platform admins"* rule in `routers/auth.py`, scoped to the **first-party** broker and killable with `MASSING_CLOUD_ROLE_SYNC=0`. The four direct IdPs in `oauth.py` keep the old rule: a Google account proves an email, not a relationship to this product.

## Admin becomes a section, not a console

The four admin-only dropdown entries — Manage users, Audit log, Errors, Data connections — are now the **Administration section** of one Profile & settings panel, present only when the account carries the capability.

That is the reorganisation asked for ("all the admin settings should be in a user's profile settings section") **without deleting a capability the product still needs**: local admin is load-bearing for user management, audit, API keys and data connections. A non-admin and an admin now see the *same* panel; the admin's has one section more. `profileSettings.test.ts` asserts exactly that — including that the admin *actions* are not built at all for a non-admin, because a hidden tab whose body still rendered would leak the buttons.

Refusals stay distinguishable throughout: a Free plan gets **402** naming the upgrade page, an unlinked account **403**, and a plan limit passes the site's own **409** message through verbatim. An empty library and an ungranted one never look the same.

## Turning it on

No secret to configure — PKCE replaces it.

```
MASSING_CLOUD_SSO_ENABLED=1
MASSING_CLOUD_SITE_URL=https://www.massing.cloud     # default
MASSING_CLOUD_SSO_CLIENT_ID=massing-desktop          # default, already registered
MASSING_CLOUD_ROLE_SYNC=1                            # default
MASSING_CLOUD_ADMIN_ROLES=administrator,editor       # default
```

Off by default; the routes 404 until enabled.

## Verification

* **Backend** — `test_massing_cloud_sso.py` drives the whole flow with the network stubbed at the module seams: PKCE properties, callback/link, role map **and role loss**, tier map, library, 402/403/409, disconnect, feature-flag-off.
* **Live HTTP** — ran the real ASGI server and confirmed all 8 routes serve, `/auth/cloud/login` 307s to the broker with an S256 challenge and a loopback `redirect_uri`, and the `mc_pkce` cookie is `HttpOnly; Path=/auth/cloud; SameSite=lax` with the verifier **absent** from the redirect URL.
* **Live UI** — drove the running app: chip renders with initials + ADMIN badge; the dropdown shows *Profile & settings…* / *Connect massing.cloud…* / *Sign out*; the panel shows all five sections with Administration's three groups.
* **Suites** — web 2005/2005, typecheck and lint clean.

Two repo gates caught real defects on the way through: `hrefGuard` flagged `avatar_url` reaching an `img.src` (now scheme-checked via `safeHref`, with a dangerous-scheme test), and a happy-dom failure that is a genuine browser concern — the space-separated `hsl(H S% L%)` form is dropped outright by some CSS parsers, which would have left the fallback disc transparent.

## Deliberately not included

* **Saving back to the cloud.** Doc 31 section 3 is an open joint decision — the site's `POST /models` records a *pointer* and assumes the bytes already exist in storage. This app has no `.mass` container writer, so a save button would record a pointer to nothing. `massing_cloud_vault.save_model` exists, unrouted.
* **Opening `.mass`** — refused explicitly rather than handed to the IFC loader.
* **Doc 31 section 4 deep links** (`massing://open?...`).

## Summary by CodeRabbit

* **New Features**
  * Added optional massing.cloud sign-in and account linking.
  * Added profile settings with identity, security, preferences, cloud, and administrator sections.
  * Added cloud project-library browsing and model opening.
  * Added account avatars with initials and reliable fallbacks.
  * Added cloud account refresh and disconnect controls.
* **Bug Fixes**
  * Improved handling of unavailable cloud services, unlinked accounts, failed avatar images, and insufficient plan access.
* **Documentation**
  * Documented massing.cloud sign-in and library integration.

### DIFF
--- a/apps/web/src/account/accountChip.test.ts
+++ b/apps/web/src/account/accountChip.test.ts
@@ -1,0 +1,142 @@
+import { beforeEach, describe, expect, it } from "vitest";
+
+import { avatarEl, initialsFor, renderChip } from "./accountChip";
+
+/**
+ * The identity chip — "signed in as", with the massing.cloud (WordPress) avatar.
+ *
+ * The interesting cases are all about the **absence** of a cloud profile, because that is the normal
+ * state for password and direct-IdP accounts and it must never read as an error. A missing
+ * `avatarUrl` has to produce a real element, and a *broken* one has to degrade to the same element
+ * rather than leaving the browser's broken-image glyph on the most-looked-at control in the shell.
+ */
+describe("initialsFor", () => {
+  it("takes one glyph from each of two name parts", () => {
+    expect(initialsFor("Ada Lovelace")).toBe("AL");
+  });
+
+  it("splits an email local-part on separators, ignoring the domain", () => {
+    // "ada.lovelace@example.com" must not become "AD" — and must never reach into the domain.
+    expect(initialsFor("ada.lovelace@example.com")).toBe("AL");
+    expect(initialsFor("ada@example.com")).toBe("AD");
+  });
+
+  it("falls back to the first two characters for a single-word name", () => {
+    expect(initialsFor("ada")).toBe("AD");
+    expect(initialsFor("x")).toBe("X");
+  });
+
+  it("never returns an empty string, so the disc is never blank", () => {
+    expect(initialsFor("")).toBe("?");
+    expect(initialsFor("   ")).toBe("?");
+    // A name that is *only* separators still has to yield something paintable.
+    expect(initialsFor("@example.com").length).toBeGreaterThan(0);
+  });
+
+  it("is at most two glyphs", () => {
+    for (const n of ["Ada Lovelace", "ada.lovelace@example.com", "Wolfgang Amadeus Mozart", "x"]) {
+      expect(initialsFor(n).length).toBeLessThanOrEqual(2);
+    }
+  });
+});
+
+describe("avatarEl", () => {
+  it("renders an <img> when the cloud supplied an avatar", () => {
+    const el = avatarEl({ username: "ada@example.com", avatarUrl: "https://cloud/a.jpg" });
+    expect(el.tagName).toBe("IMG");
+    expect((el as HTMLImageElement).src).toBe("https://cloud/a.jpg");
+    // Decorative: the accessible name is on the button, not repeated here.
+    expect((el as HTMLImageElement).alt).toBe("");
+    // Never leak the app's URL to the avatar host as a referrer.
+    expect((el as HTMLImageElement).referrerPolicy).toBe("no-referrer");
+  });
+
+  it("refuses a dangerous scheme and falls back to the disc", () => {
+    // `avatar_url` is external data — it comes from massing.cloud's userinfo, which reflects
+    // whatever avatar host the account uses. A non-http(s) scheme must never reach `src`.
+    for (const bad of ["javascript:alert(1)", "data:text/html,<script>", "vbscript:x", "file:///etc"]) {
+      const el = avatarEl({ username: "ada@example.com", displayName: "Ada Lovelace", avatarUrl: bad });
+      expect(el.tagName, `${bad} must not become an <img>`).toBe("SPAN");
+      expect(el.textContent).toBe("AL");
+    }
+  });
+
+  it("still accepts an ordinary https avatar", () => {
+    const el = avatarEl({ username: "ada@example.com", avatarUrl: "https://www.massing.cloud/a.jpg" });
+    expect(el.tagName).toBe("IMG");
+  });
+
+  it("renders an initials disc when there is no avatar", () => {
+    const el = avatarEl({ username: "ada@example.com", displayName: "Ada Lovelace" });
+    expect(el.tagName).toBe("SPAN");
+    expect(el.textContent).toBe("AL");
+    expect(el.getAttribute("aria-hidden")).toBe("true");
+  });
+
+  it("swaps a broken avatar for the disc instead of showing a broken image", () => {
+    const el = avatarEl({ username: "ada@example.com", displayName: "Ada Lovelace",
+      avatarUrl: "https://cloud/gone.jpg" }) as HTMLImageElement;
+    const host = document.createElement("div");
+    host.append(el);
+    el.onerror?.(new Event("error"));
+    expect(host.querySelector("img")).toBeNull();
+    expect(host.textContent).toBe("AL");
+  });
+
+  it("gives the same person the same fallback colour across renders", () => {
+    // Read the raw style attribute, NOT `el.style.background`: happy-dom does not reflect the
+    // `background` shorthand back, so both sides would be "" and the equality assertion below
+    // would pass no matter what the code did.
+    const hue = (el: HTMLElement) => {
+      const css = el.getAttribute("style") || "";
+      const m = /hsl\((\d+)/.exec(css);
+      expect(m, `expected an hsl() background in: ${css}`).not.toBeNull();
+      return m![1];
+    };
+    expect(hue(avatarEl({ username: "ada@example.com" })))
+      .toBe(hue(avatarEl({ username: "ada@example.com" })));
+    expect(hue(avatarEl({ username: "ada@example.com" })))
+      .not.toBe(hue(avatarEl({ username: "bob@example.com" })));
+  });
+});
+
+describe("renderChip", () => {
+  let btn: HTMLButtonElement;
+  beforeEach(() => { btn = document.createElement("button"); });
+
+  it("shows the display name, not the raw username, when the cloud gave one", () => {
+    renderChip(btn, { username: "ada@example.com", displayName: "Ada Lovelace" });
+    expect(btn.textContent).toContain("Ada Lovelace");
+    expect(btn.getAttribute("aria-label")).toBe("Account: Ada Lovelace");
+  });
+
+  it("falls back to the username when there is no display name", () => {
+    renderChip(btn, { username: "ada@example.com" });
+    expect(btn.textContent).toContain("ada@example.com");
+    expect(btn.getAttribute("aria-label")).toBe("Account: ada@example.com");
+  });
+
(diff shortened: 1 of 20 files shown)


ibuilder:
### Status update — the gates that were failing are fixed

This branch was briefly on `main` by accident (a concurrent session in the same clone moved `main` forward with `git branch -f` and my commit sat between its two releases), and was correctly [reverted off main](https://github.com/ibuilder/massing/commit/a3026994). **Nothing needed recovering** — and that revert's message named three failing gates I had not yet found, because my own suite runs were racing the other session's and never printed a clean list.

All four are now fixed on this branch:

**1. `test_alembic_migrations`** — the model declared `linked_at: Mapped[datetime]` (NOT NULL) while the migration wrote `nullable=True`. Real drift; the two would have disagreed on a fresh Postgres. Migration now `nullable=False`, matching the baseline's other non-optional timestamps.

**2. `test_global_authz`** — `POST /auth/cloud/refresh` and `POST /auth/cloud/disconnect` carried only `Depends(current_user)`, which **identifies but does not authorise**: with RBAC on and no credential it returns the literal string `"anonymous"`. Both now take `require_identified`, as do the three `/cloud/library` reads.

`GET /auth/cloud/status` deliberately keeps `current_user`, and the docstring says why: it is the one cloud route an **anonymous caller must reach**, because the sign-in modal calls it while signed out to decide whether to offer "Continue with massing.cloud". Gating it would 401 exactly the caller it exists to serve. An anonymous caller gets `{enabled, linked: false, site_url}` — deployment configuration, no user data, never a token.

**3. `test_protected_prefix_coverage`** — `/cloud` was a new unclassified top-level prefix, now in `_PROTECTED_PREFIXES` (every route under it reads one user's vault — tenant state). `/auth/cloud/*` is deliberately *not* covered, or signing in would require being signed in. The middleware is defence in depth and **not** the gate: the routes each carry their own `require_identified`, because an armed middleware makes a guarded and an unguarded route look identical from outside.

**4. `test_route_reachability`** — `/auth/cloud/callback` has no web caller and **must not have one**: it is the OAuth redirect target, so its caller is the *broker*. Added to `KNOWN_UNCALLED` with that reason and no expiry condition. Its pair `/auth/cloud/login` does have a caller (`cloudLoginUrl`), which is what proves the flow is wired — a grep of web source structurally cannot see the other half of a redirect.

Verified live with RBAC on, anonymous:

```
GET  /auth/cloud/status      200   (config only — the sign-in button needs it)
GET  /auth/cloud/login       307
POST /auth/cloud/refresh     403
POST /auth/cloud/disconnect  403
GET  /cloud/library/projects 401
```

To land: this branch, then revert the revert.

ibuilder:
### Green, and the PR now actually carries the feature

**Backend 625/626** (2060s, 15-way parallel) · **Web 2006/2006** · typecheck + lint clean.

The one backend failure is `test_clash_federated_job`, which **passes 3/3 in isolation** and shares no code with this branch (`grep -c 'cloud|CloudIdentity|require_identified|massing_cloud'` over that test → 0). It is a parallel-load flake in the job-queue suite. Sweep residue on this run was clean — 0 unaccounted databases — so unlike my earlier runs this one was not poisoned by a concurrent session.

### The merge, and why it needed a content check rather than a presence check

`main`'s only commit ahead of this branch was [the revert](https://github.com/ibuilder/massing/commit/a3026994). Because `7edb59f5` is now an **ancestor of main**, `git log origin/main..HEAD` returned only the fix commits — this PR would have merged green and landed nothing. So the un-revert rides inside it as one atomic change.

Merging surfaced four modify/delete conflicts, which were the harmless ones. The damage was silent:

* **Eight files staged for deletion with no conflict at all** — nothing conflicts when one side is a deletion and this branch had not touched them since.
* **`models.py` auto-merged cleanly and dropped the `CloudIdentity` table outright.**

`git ls-tree` would have reported all twenty paths present in that broken state. The check that discriminates is content, and it is now empty:

```
git diff 96fe3bf2 HEAD -- <the 20 reverted paths>   # → empty
```

Resolution was a full restore of all twenty paths from the pre-merge tip, not a fix of the four visible conflicts. The path list is not hand-written: it is `git show --name-only a3026994`, asserted identical to the feature commit's own set.

PR diff is now **21 files, +2362** — the feature is really in the range.

### Gates re-run on the merged tree specifically

A gate that passed pre-merge was a statement about a tree that will never be `main`. On the merged tree: `test_alembic_migrations`, `test_alembic_single_head`, `test_global_authz`, `test_protected_prefix_coverage`, `test_route_reachability`, `test_dead_code_population`, `test_massing_cloud_sso` — all pass. The migration fix and the migration itself exist together for the first time on this tree, which is the only tree on which `alembic check` means anything.

### Two more defects found since the last update

* **`test_dead_code_population`** — `save_model` / `delete_model` were public functions with no caller anywhere. **Removed, not allowlisted**: the byte-upload half is an open joint decision (docs/31 §3) and there is no `.mass` writer to feed them. That gate's population was corrected 877 → 35 → 13 with eight survivors turning out to be live symbols, so it is not a rule to argue with from the comfort of having just written the code.
* **`docsPublished.test.ts`** — the new internal note was not in `docs/internal/README.md`. Missed because the web run I earlier cited as green **predated the document's creation**, so the gate was never given a chance to fail.

### Still true, and still the one thing this cannot do alone

`userinfo` publishes **no role** — verified against `class-broker.php`, not the docs. The admin/editor mapping is built, two-way, fail-closed, and **inert** until massing.cloud adds it. The four-line `massing_sso_userinfo` filter is in `docs/internal/massing-cloud-sso.md`. And the real broker handshake has never been exercised: every test stubs it at the module seam.

ibuilder:
⚠️ **This PR's "Merged" badge is wrong — nothing from it is on `main`.**

GitHub auto-closed it as merged when a concurrent session in the same clone pushed the commit directly to `main` (it moved `main` with `git branch -f` while HEAD sat on this branch). That commit was then correctly [reverted off main](https://github.com/ibuilder/massing/commit/a3026994) because it turned `main` red.

Current state of `main`:

```
git ls-tree -r --name-only origin/main | grep -cE 'massing_cloud_auth|routers/cloud.py|account/accountChip'
# → 0
```

Superseded by #338, which carries the feature, the four gate fixes and the un-revert as one atomic change. Do not treat this PR as delivered.

PR #344, opened 2026-08-26
v0.3.1101–1105 — R22-REPORT-BUILDER, all four remaining items

ibuilder:
# What & why

The first roadmap feature work of this session, after four PRs of infrastructure. The sprint table's row 1 (a correctness sweep) was struck in #340, making **row 2** next — and that row carried a premise-check.

## The premise-check found the row itself was wrong

The row read **"items 2–4"**, premise: *"That three items really remain. v0.3.1088 shipped a part… Re-read the entry, not the release note."*

The entry lists **five** items. One done, **four** remaining.

The cause is recorded inside the entry: item 5 *"was missing from the four-item list above for a day"*. It was added later; the sprint row was written in the four-item era and never re-derived. **The exact failure that row's own premise-check existed to catch, sitting in the row that asked for it.**

### Item 3 — a config schema (v0.3.1101), and the live miscount it exposed

`SavedView.config` was stored verbatim. `validate_view_config` resolves every field through `_resolve_field` and every operator through `FILTER_OPS` — **the same two the list route uses**, because `_json_text` interpolates a field name into a JSON path, so an unvalidated name is an injection site and a second validator would be a second answer to "what is a field". Unknown keys are **refused**, which is what gives the schema a migration path.

Then the payoff. `count_records` has taken `filters` since MOD-FILTER, and its docstring names this caller — *"a count that ignores a filter the list applied reports a total the page cannot account for."* **`view_alerts` never passed it.** Measured on two RFIs differing by one declared field:

```
alert feed total = 2        the view itself shows = 1
```

No browser produced it (the register saves only `{q,state,sort}`), but the API accepted filters from anything else and the schema now *invites* them — so the two had to land together.

### Item 4 — shareable views (v0.3.1102)

`scope` separates **ownership** (who may write) from **visibility** (who may read). Asserted three ways: not across projects, not into edit rights, and **not into the alert feed** — a deliberate non-feature, because `last_seen_at` is one column per row and a shared view in a second person's feed would compute "new since" from *the author's* last visit. Per-viewer alerts need a per-viewer timestamp; the gap is pinned rather than assumed.

### Item 2 — cross-module reports (v0.3.1103)

`aggregate` takes a `join` **resolved against `reference_fields`, never accepted from the request**. The join is **LEFT** deliberately: an inner join silently drops every base record without a match — mutation-verified, it turns the fixture's **4 of 4 into 3 of 4** while looking complete.

The test picks its modules **by resolution**, walking the live registry for a real declared edge (`action_item.responsible_contact → contact`). A fixture naming modules by hand keeps passing after the schema connecting them changes.

### Item 5 — one report surface (v0.3.1104)

The entry said the *decision* was the deliverable. **They stay separate implementations behind one surface.** A built-in report is **code** (EVM, WIP, appraisal — 56 of them) computing what no query builder expresses; a saved view is **data**. Folding either into the other buys a query language that can do EVM, or 56 rows of config that secretly dispatch to Python.

The duplication was never the implementations — it was the **surface**. `GET /reports` was global and knew nothing about saved views, so one question got two answers from two panels. `GET /projects/{pid}/reports/catalog` is the one answer.

## Three things I got wrong, all caught before landing

| what | how it was caught |
|---|---|
| The "uncountable" guard in `view_alerts` was **dead code** — `view_filters` was tolerant by shape, so an unreadable config returned `[]` and got counted as unfiltered: the same wrong number, by the guard meant to prevent it | the new test, on its first run |
| A sharing assertion was a **tautology** (`X or True`) over a fixture where Bob owned his own row of that name — a name-based check couldn't have distinguished a leak either way | reading my own diff |
| My roadmap-archiving regex **cut four unrelated items**, because the next entry starts `- ◧ **` and my pattern required `- **` | the roadmap gates |

## And the schema's first act was to catch a key nothing had ever read

`test_view_delete` saved views as `{"filter": name}` — **singular**. Every reader in the repository looks for `filters`/`q`/`state`/`sort`; the only writer of `filter` was that one fixture line. **A key written and read by nobody, invisible for exactly as long as configs were stored verbatim** — which is the defect item 3 describes, caught in the act against the repo's own suite.

That failure also arrived misreported: the detail string printed `len(VIEW_CONFIG_KEYS)` — the *expectation* — so it read "10 keys round-tripped" while 9 had. It now prints the actual count and names the missing key.

**Both failures came from the full suite, not my targeted runs**, which had passed because they predated adding `join` to the key set.

## Anti-vacuity floors re-seated, not shaved

Archiving the item dropped three floors below their minimums (25→15, 28→15, 10→5). Set from **what makes them non-vacuous**, not from today's count — which is the rule `roadmapSelfConsistent.test.ts` already states about exactly this situation: *"a floor tracking a deliberately shrinking population must be set from what makes it non-vacuous, never from what the file happens to contain today."* Shaving by one is the ratchet-in-reverse it warns about.

- [x] Backend: `ruff` clean; **635/635 suites pass**, 0 kept DBs, clean sweep
- [x] Four new tests registered in `run_tests.py`; manifest guard shows no drift
- [x] Migration `c1a4e7b28d61` — single head confirmed (24 migrations)
- [x] Web: **2,010 tests / 197 files**; `tsc --noEmit` clean; root + web ESLint clean; fragments/Node-base guard green
- [x] `CHANGELOG.md` entries added (newest at top); roadmap entry archived to `roadmap-completed.md`
- [x] Version bumped in all three files → 0.3.1104
- [x] No secrets; no competitor names

## One thing for you

Item 4 ships sharing but **not per-viewer alerts** — a shared view only alerts its author. That is a deliberate stopping point, not an oversight: doing it properly needs a per-viewer `last_seen_at` table, and faking it would reproduce the exact miscount item 3 just removed. Flagged here so the gap is yours to schedule rather than mine to assume.

---

## Summary by CodeRabbit

* **New Features**
  * Added project sharing for saved views, with private and project-wide visibility options.
  * Added unified project report catalogs combining built-in and saved-view reports.
  * Added cross-module report aggregation with validated left joins.
  * Added validation for saved-view configurations and filters.

* **Bug Fixes**
  * Alert counts now respect saved-view filters and clearly identify unreadable configurations.
  * Improved ownership and visibility boundaries for shared views.

* **Documentation**
  * Updated changelog and roadmap records for the completed reporting enhancements.

* **Chores**
  * Updated the application version to 0.3.1104.


ibuilder:
Both merge-risk findings were real. Fixed in `9b8087f` (v0.3.1105).

**1. "Invalid aggregate configurations can still be saved and later fail when a report is replayed."**

Confirmed, and `validate_view_config`'s own docstring was the evidence against it — it promises *"a view is validated exactly as the query it replays would be — otherwise a view could be saved that the aggregate route then refuses,"* while the code kept that promise for field **names** only. `aggregate()` enforces two rules the validator never asked: a non-`count` aggregate needs an `agg_field`, and `sum`/`avg` need a numeric one. Measured before the fix:

```
{"group_by":"target","agg":"sum"}                        -> SAVED OK, 400 on replay
{"group_by":"target","agg":"sum","agg_field":"target"}   -> SAVED OK, 400 on replay
{"group_by":"target","agg":"avg","agg_field":"target"}   -> SAVED OK, 400 on replay
```

Both rules are now re-asked of the *same resolved field definition* `aggregate` uses rather than re-derived from the name — a second answer to "is this field numeric" is how two sources of truth start disagreeing.

Covered in `test_view_config.py`, both halves: `rfi` declares no numeric field, so the acceptance cases could not be expressed against it, and refusals alone would be satisfied by a validator that refuses *every* aggregate. The numeric case resolves against whichever module declares a numeric field, guarded by an assertion that one was found. Mutation-checked — with the new block neutered, the three refusal cases fail and the three acceptance cases still pass.

**2. "The archived roadmap description does not match the shipped cross-module behavior."**

Also correct. Item 2 was the one of five that never got its "landed" line, so it still read as an open problem statement after shipping in v0.3.1103. Corrected — and the correction records the part that is easy to misread: `SavedView.module` **is** still one string. That is the design rather than a shortfall; the base register is singular and the second module arrives as a declared-reference `join`, which is what keeps the edge coming from the schema instead of from the request.

On the docstring-coverage warning: not addressed. The functions it counts carry block comments explaining *why* rather than `"""` docstrings, and adding docstrings to hit a percentage would be writing for the metric.

---

PR #384, opened 2026-08-31
v0.3.1127 — the roadmap described a defect that a release had already ended

ibuilder:
# What & why

Two `docs/roadmap.md` entries — one in the R46 block, one in the open-decisions list — claimed that `eot.py` **"names four AACE methods and performs none of them. All four return an identical number on the same input."** Both are corrected. The claim is false, and it was false on the day it was written.

Measured through `eot.analyse` on one job (baseline finish 2026-01-31, two excusable events totalling 20 days):

| actual finish | `impacted_as_planned` | `as_planned_vs_as_built` | `time_impact` | `windows` |
|---|---|---|---|---|
| 2026-02-10 (slip 10) | **20.0** | **10.0** *(capped)* | *refuses* | *refuses* |
| 2026-03-12 (slip 40) | 20.0 | 20.0 *(cap idle)* | *refuses* | *refuses* |

Three distinct outcomes where the as-built cap binds, two where it does not — and **the two series methods never return a number at all**, refusing with `method_needs_schedule_updates`. So "all four return an identical number" is untrue on *every* input, not merely on some.

## The fix predated the claim by four days

`v0.3.971` shipped **2026-08-16** under the title *"four delay methods, one number, and a label that said otherwise"* — it is the release that ended exactly this. The roadmap paragraph was dated *"measured 2026-08-20"*.

That date is the reason it survived. **A measurement date is what stops a reader re-checking a sentence**, so inheriting one made this harder to catch, not easier.

## Its own citation contradicted it

The paragraph cited `services/api/test_schedule_windows.py` as its pin. That test asserts `windows` and `time_impact` return `None` — *"which is now the REASON they refuse"*. The gate had been green on the **corrected** behaviour the entire time.

*A citation is evidence that a file exists, not that it says what the sentence claims.* `test_claude_md_gates.py` checks the former by design and structurally cannot check the latter.

## No new gate, deliberately

The behaviour is already pinned twice — `test_schedule_windows.py` and `test_eot_methods.py` — and both were green throughout. What failed was prose that the existing gates already contradicted, not a missing check. A string-matching guard on one sentence would rot the same way the sentence did.

## What survives

The genuine open decision is preserved and unchanged: **whether `/schedule/eot` should delegate to `schedule_windows` and `schedule_modelled` — which perform three of the four for real — or keep its own number and cite theirs.** An EOT figure ends up in arbitration, so that stays a domain call, not a refactor.

The separate "EOT surface is dark" item was also re-verified and is **accurate as written**: both routes are live (`main.py:559`), `grep "schedule/eot" apps/web/src` returns 0, `/schedule/eot/sourced` is recorded in `KNOWN_UNCALLED`, and `/schedule/eot` escapes that gate only because its leaf `eot` is 3 characters against `MIN_SEGMENT = 5`. Left untouched — it is a scope decision, not a defect.

- [x] Backend: `cd services/api && python -m ruff check src/ ../data/src/` clean
- [x] Backend: affected `test_*.py` pass — `test_eot`, `test_eot_methods`, `test_eot_sourced`, `test_claude_md_gates`, `test_roadmap_status`, `test_doc_substance`, `test_file_sizes`. No new test files, so the `run_tests.py` TESTS manifest is unchanged
- [x] Web: typecheck clean; full suite **2045 tests / 200 files**, including the three roadmap parsers (`roadmapStale`, `roadmapSelfConsistent`, `roadmapLanes` — 33 tests) that are sensitive to bullet shape
- [x] No import cycles introduced (docs-only change; no source file touched)
- [x] `CHANGELOG.md` entry added (newest at top); both roadmap entries corrected
- [x] Version bumped in `apps/web/package.json`, `apps/web/src-tauri/tauri.conf.json` **and** `package-lock.json` → 0.3.1127

**Docs only — no engine changed.** The two surviving occurrences of the false wording are inside the quoted-and-corrected record, which is intentional: deleting the wrong sentence would destroy the evidence of what was wrong.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,43 @@ All notable changes to Massing. Releases are signed, auto-updating desktop build
 (Windows / macOS / Linux); the updater always serves the latest. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/).
 
+## v0.3.1127 (2026-08-31) — the roadmap described a defect that a release had already ended
+
+Documentation only; no engine changed. Two roadmap entries said `eot.py` **"names four AACE methods
+and performs none of them. All four return an identical number on the same input."** Both are
+corrected, because the claim was false — and false on the day it was written.
+
+Measured through `eot.analyse`, one job, baseline finish 2026-01-31, two excusable events totalling
+20 days:
+
+| actual finish | `impacted_as_planned` | `as_planned_vs_as_built` | `time_impact` | `windows` |
+|---|---|---|---|---|
+| 2026-02-10 (slip 10) | **20.0** | **10.0** *(capped)* | *refuses* | *refuses* |
+| 2026-03-12 (slip 40) | 20.0 | 20.0 *(cap idle)* | *refuses* | *refuses* |
+
+Three distinct outcomes where the as-built cap binds, two where it does not, and the two series
+methods never return a number at all — they refuse with `method_needs_schedule_updates`. "All four
+return an identical number" is therefore untrue on every input, not merely on some.
+
+**The fix predated the claim by four days.** `v0.3.971` shipped 2026-08-16 under the title *"four
+delay methods, one number, and a label that said otherwise"*; the roadmap paragraph was dated
+*"measured 2026-08-20"*. A measurement date is what stops a reader re-checking a sentence, so an
+inherited date made this harder to catch, not easier.
+
+**The citation contradicted the claim it supported.** The paragraph cited
+`services/api/test_schedule_windows.py` as its pin. That test asserts `windows` and `time_impact`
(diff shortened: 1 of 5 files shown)


ibuilder:
@coderabbitai review

(The rate limit was hit while this was still a draft, and the retry doesn't appear to have fired after the reset. CI is green on `b7bac0b4` — all 14 checks, `API test gate` included.)

This one is documentation only, no engine changed, so the useful review surface is whether the corrected prose is internally consistent and whether it now says the true thing. Worth noting what it replaces: the claim being corrected was confidently worded and carried its own measurement date, which is exactly what stopped anyone re-checking it for eleven days.

---

PR #387, opened 2026-08-31
v0.3.1130 — a release you could sign and nobody could check

ibuilder:
# What & why

`ASSET-VERIFY`. `asset_rights.py` shipped both halves of a signed release manifest and **only sealing was reachable**. Counted against `services/api/src`, `services/data/src` and `apps/web/src`, excluding the module itself:

| function | production callers |
|---|---|
| `verify_release` · `verify_signature` · `verify_content_hash` · `verify_manifest_hash` | **0** |
| `public_key_b64` · `generate_seed` | **0** |
| `sign_manifest` · `build_manifest` · `new_asset_id` | 1 each |

A manifest exists so that *someone else* can confirm a release is authentic and unaltered. Sealing alone produces a file nobody can check — including us.

* **`POST /asset-rights/verify`** takes the `asset_rights.json` out of a `.mass` and reports findings separately — `content_hash_ok`, `manifest_hash_ok`, `signed`, `signature_ok`, `trusted_key` — because *"the content was altered"* and *"it carries no signature"* call for different responses.
* **`GET /asset-rights/status` now serves `public_key`.** Its docstring said *"never the key itself"*, which read as a rule against publishing **any** key. The public half is what verifiers need, and the same key is already embedded in every signed manifest we emit — withholding it protected nothing and left a recipient trusting the document's own copy, which is the copy an attacker would have replaced.
* **`python -m aec_api.asset_rights --generate`** mints a signing seed. Without it the *signed* path was unreachable from a clean deployment except by generating an Ed25519 key out of band. A command and deliberately **not** a route: minting a private key is an operator action at the machine.

## The design error, caught by writing the test rather than reading the code

The plan was to default `public_key` to this deployment's key "so `trusted_key` means something". That is backwards, and the test's trust matrix is what showed it:

| supplied key | `signature_ok` | `trusted_key` |
|---|---|---|
| none — falls back to the key **in the document** | true | **false** |
| the right trusted key | true | **true** |
| a *different* key | **false** | false |

A release signed by anyone else would land in that third row — reported as a **bad signature** when it is perfectly valid. The key is the caller's to supply.

## A client method was written and then deleted

`verifyRelease()` had nothing in the UI calling it, so it would have been an **unwired client method added to make a reachability gate go green** — the exact defect this release removes, introduced while removing it. The party who verifies a release is whoever received the file, with their own tooling, like `/auth/cloud/callback`.

## A blind spot recorded rather than worked around

The reachability gate **cannot see the new route in either direction**: the leaf `verify` occurs in 25 unrelated web files, so the substring rule reads it as called — and for the same reason it cannot be frozen as uncalled. Asserted in `test_route_reachability.py` beside its other documented blind spot, with measured alternatives for whoever takes the matcher on: `/leaf` flags **43** further routes, a word-boundary match flags **5**. Neither belongs in a release about signed manifests, and each needs those routes triaged rather than bulk-frozen.

- [x] Backend: `cd services/api && python -m ruff check src/ ../data/src/` clean
- [x] Backend: new gate `test_asset_verify.py` (20 checks, 4 mutations) **registered in the `run_tests.py` TESTS manifest**; `test_tested_but_unwired` frozen entries removed for the now-wired functions; `test_route_reachability`, `test_reachable`, `test_import_cycles`, `test_claude_md_gates`, `test_roadmap_status`, `test_doc_substance`, `test_file_sizes`, `test_declared_imports` all green. **Full backend suite running; this PR will not be merged until it is green.**
- [x] Web: typecheck clean; full suite **2045 tests / 200 files**, including the four roadmap parsers
- [x] `CHANGELOG.md` entry added (newest at top); roadmap entry closed ✅ and removed from the Lane C row
- [x] Version bumped in `apps/web/package.json`, `apps/web/src-tauri/tauri.conf.json` **and** `package-lock.json` → 0.3.1130
- [x] No secrets, no competitor names in shipped docs — and the seed never reaches the API surface or the `--generate` guidance text, both asserted

One mutation found a hole in the test itself: without the `signing_available()` guard, `--public-key` **crashes** rather than refuses, and an exit-code assertion cannot tell those apart. What the guard buys is an operator being told what to set, so that is what is asserted now.

---

## Summary by CodeRabbit

* **New Features**
  * Added asset-rights release manifest verification for identified callers.
  * Asset-rights status now exposes the public signing key.
  * Added operator commands to generate signing seeds and display public keys.

* **Bug Fixes**
  * Improved handling and reporting of invalid signing keys and verification findings.
  * Prevented unauthenticated access to manifest verification.

* **Documentation**
  * Documented the ASSET-VERIFY release and updated the roadmap.

* **Chores**
  * Updated the application version to 0.3.1130.
  * Added coverage for verification workflows and route availability.


ibuilder:
Both risks were real. Fixed in `27ddc773` — and the first one turned out bigger than the feature.

## 1. Malformed signing key — confirmed, and worse than described

Reproduced before touching anything, with `AEC_ASSET_SIGNING_KEY=not-a-real-seed`:

```
signing_available()          -> True          # it only asked: is the variable non-empty?
GET /asset-rights/status     -> unhandled exception, not a clean 500
sign_manifest(...)           -> RuntimeError: must decode to exactly 32 bytes, got 11
```

That third line is the important one. **`bundle.py` signs whenever `signing_available()` is true**, so a malformed key was already taking down the *entire `.mass` export* — that half pre-dates this PR. Publishing the key on `/asset-rights/status` inherited the same fault and made it a crash on the one route a client calls to decide whether to *offer* sealing; that half is a regression this PR introduced, and you caught it.

The root cause is the one this whole series keeps finding: **`signing_available()` reported a claim, not a fact.** It now decodes the seed and builds the key, so a misconfigured deployment reports `signing: false` — hashed, tamper-evident, unsigned, a state the contract already has words for — instead of failing downstream where the cause is invisible:

```
malformed -> signing_available=False, status 200 {"signing": false, "public_key": ""}
unset     -> signing_available=False
valid     -> signing_available=True   (positive control)
```

The CLI also now distinguishes *unset* from *set but unusable*: telling an operator their key is missing when they can see it in the environment just sends them to re-set the same typo. Asserted over three malformed shapes (non-base64, junk, valid-base64-but-too-short) and mutation-checked both ways.

## 2. `trusted_key` — real, and **not fixable as stated**, so it is documented and pinned

You are right that it can mislead. The sharp version is worse than the general one:

```
same manifest, key echoed from verification.public_key -> trusted_key: True
same manifest, no key supplied                          -> trusted_key: False
```

Identical evidence, opposite answers, decided by whether the caller copied a value out of the file.

But the implied fix — comparing the supplied key against the embedded one — would be **wrong**: for a genuine release the trusted key *is* the embedded one, so that check would reject exactly the case it should accept. There is no trust anchor the API can consult; trusting a key is the one part a verifier must do out of band, from the issuer's published key or a channel independent of the file.

So I documented the footgun on the route and **asserted the flip in the test**, so it reads as a known limit rather than an undiscovered bug — and asserted that `signature_ok`, the part actually proven, is identical either way.

I'd rather state a boundary than add a field that implies the API validates trust it cannot see.

---

PR #389, opened 2026-08-31
v0.3.1132 — eight routes frozen as callerless while the UI called them

ibuilder:
# What & why

`ROUTE-INTERP`. `test_route_reachability` flags a route when its **last static segment** appears nowhere in the web source. The file's header already knew literal matching is unsound against a client that builds URLs from template literals — that is *why* the rule is the last segment. **It assumes the last segment is written out.**

Eight routes template it too, so their leaf appears nowhere, and every one was frozen in `KNOWN_UNCALLED` as deliberately callerless. Each is listed as *URL built at → value passed at*, because the second is the evidence of reachability and the first alone is not:

| route | URL built at | value passed at |
|---|---|---|
| `/exports/{cobie,qto,schedule,spaces}.xlsx` | `viewer/tools/exportsSection.ts:57` | the four-entry list on line 54 |
| `/schedule/{gantt,lob}.svg` | `api/schedule.ts:193-4` | `portal/panels/schedule.ts:731` |
| `/model/export.{jsonld,parquet}` | `api/model.ts:185` | `portal/panels/standards.ts:488-489` |

The list's own contract is that *"a frozen route that gains a caller must leave"*. These never lacked one. Four sat under a heading reading *"a user reaches these by clicking a link the server builds, so no client caller is expected"* — **which is what made them look considered.**

## Two mechanisms, because one shape decides and the other cannot

**Templated stem** (`` `exports/${file}.xlsx` ``) is a **pattern**: parent segment and extension must still match literally. Dropping the parent anchor *would* vouch for six further routes nothing calls — measured as a mutation, not shipped behaviour.

**Templated extension** (`` `model/export.${fmt}` ``) is a **two-entry named list with its call site recorded**, because as a pattern it is wrong in *both* directions — it vouches for six `export.*` routes of which only two are called. The four that stay frozen each say the same thing, and it is the finding worth keeping: **a signature that ACCEPTS a value is not a call site that PASSES one.**

| route | why it stays frozen |
|---|---|
| `/model/export.ifcx` | `modelExportUrl` is typed `"csv" \| "jsonld" \| "parquet"` |
| `/drawings/sheet.dxf` | `sheetPath` accepts `"dxf"`; its only caller `openSheet` is `("svg" \| "pdf")` |
| `/energy/export.{gbxml,idf}` | **`energyExportUrl` has no caller at all** — an unwired client method |

That last row is a **new finding this release does not fix**. Recorded in the roadmap rather than bundled in here.

## The leniency is pinned, because it moves the number the *rewarded* way

Every previous correction to this file made the gate stricter. This one makes it more permissive, and the file's own generated-types comment says that is the direction nobody investigates. So it is not trusted to be small:

- measured against the pre-change rule, with **every route it vouches for named**;
- **attributed per mechanism** — six by pattern, two by list. Swapping them keeps the total right and fails the attribution;
- the named list gets the mirror of `KNOWN_UNCALLED`'s rot check. An allowlist asserting *"this IS called"* outlives its caller exactly the way these eight outlived nothing.

**Five mutations, all failing, all restored:** leniency off (the six resurface) · extension-templating allowed (`sheet.dxf` + 3 more vouched for) · parent anchor dropped (six more) · an evidence string broken (rot check fires) · a stem route moved into the named list (attribution fires).

**This is not the matcher-sharpening this file has rejected twice.** That proposal went the other way — `/leaf` instead of substring, measured at 81 → 124 frozen, 43 newly unreachable. Opposite change to the same rule.

Uncalled by the rule: **81 → 73** — the strict baseline is 81 and the two leniencies remove eight.

## Three wrong turns, all in the same direction

None was caught by refining the pattern:

1. a leaf-grep of the **web source alone** — but the category's claim is a link the *server* builds, so that was the wrong corpus;
2. adding the server, a bare-leaf grep read `Content-Disposition` filenames and zip member names as links;
3. a full-path grep said 18/21 unlinked — wrong, because the URL is templated.

Reading `exportsSection.ts` and `standards.ts` is what corrected each one. *A completeness verdict computed over a pattern is confident and unfounded; the call site decides.*

## From review

All three CodeRabbit findings addressed in `94813662`, one of them a real factual error (the count was `81 → 75`; it is `73` — 75 was the six-route number, carried forward when the release grew to eight and never re-measured).

Its **nitpick was the most valuable comment on the PR**: `strip_comments(BLOB)` sat inside the `_STRICT` set comprehension, re-stripping a 3.9 MB blob once per candidate route — 943 times — making this the slowest gate in the 659-suite run. Fixed in `1110b70f`, **70.4s → 8.9s**, behaviour unchanged and the new pins still fail under mutation. The correctness half matters more than the speed: all four checks now compare against the same stripped text, which repeated calls only happened to do.

## ⛔ Blocked on a release tag, not on this diff

`API test gate` is red, **658/659**, and the one failure is `test_release_current`: *"v0.3.1132 is bumped but v0.3.1121 is the newest TAG — 11 versions have been merged and never released… Tag the release rather than raising this bound"*. Reproduced identically locally, so not a flake.

Not fixed here on purpose: the remedy is a `v*` tag, which triggers `desktop.yml` to cut a GitHub Release and upload the binaries the auto-updater serves to users. That is a publish and the owner's call; raising the bound would be switching off the check that noticed. Raised with them — see the comment thread.

- [x] Backend: `python -m ruff check src/ ../data/src/` clean
- [x] Backend: full suite **658/659** locally and in CI, the single failure being the release-tag gate above
- [x] Web: typecheck clean; `eslint --max-warnings=0` clean; full suite **2048 tests / 201 files**
- [x] No import cycles introduced — the change is one test file plus docs
- [x] `CHANGELOG.md` entry added (newest at top); roadmap entry closed ✅ under Band 2
- [x] Version bumped in `apps/web/package.json`, `apps/web/src-tauri/tauri.conf.json` **and** `package-lock.json` → 0.3.1132

**Deliberately not in this PR:** wiring `energyExportUrl`; and the larger disagreement it exposed — Band 2 names three remaining "built but unreachable" items while the gate freezes **73** routes, 38 under no category at all and 7 explicitly labelled *"worth someone's attention"*.

## Summary by CodeRabbit

* **New Features**
  * Improved route reachability detection for templated URLs and extensions.
  * Eight previously flagged routes are now correctly recognized as reachable.
* **Bug Fixes**
  * Preserved accurate detection for routes without verified client call sites.
  * Reduced false or stale route evidence.
* **Documentation**
  * Updated the changelog and roadmap with route analysis details and clarified hypothetical behavior.
* **Release**
  * Updated the application version to 0.3.1132.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,8 +26,8 @@ caller is expected"*, which is what made them look considered.
 ## Two mechanisms, because one shape is decidable and the other is not
 
 * **Templated stem** (`exports/${file}.xlsx`) is a **pattern**: the parent segment and the extension
-  must still match literally. Six routes, and dropping the parent anchor vouches for six more that
-  nothing calls.
+  must still match literally. Six routes are accepted. Dropping the parent anchor *would* vouch for
+  six more that nothing calls — measured as a mutation, not shipped behaviour.
 * **Templated extension** (`model/export.${fmt}`) is a **two-entry named list with its call site
   recorded**, because as a pattern it is wrong in both directions: it vouches for six `export.*`
   routes of which only two are called. The other four are frozen correctly, and each says the same
@@ -56,7 +56,7 @@ nothing. Five mutations, all failing, all restored.
 way — `/leaf` instead of substring, measured at 81 → 124 frozen, 43 newly unreachable. Opposite
 change to the same rule; the earlier decision does not speak to this one.
 
-Uncalled by the rule: **81 → 75**.
+Uncalled by the rule: **81 → 73** — the strict baseline is 81 and the two leniencies remove eight.
 
 ## v0.3.1131 (2026-08-31) — nine comments describing something else, and the reason nobody widened the gate
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,60 @@ All notable changes to Massing. Releases are signed, auto-updating desktop build
 (Windows / macOS / Linux); the updater always serves the latest. Format loosely follows
 [Keep a Changelog](https://keepachangelog.com/).
 
+## v0.3.1132 (2026-08-31) — eight routes frozen as callerless while the UI called them
+
+`ROUTE-INTERP`. `test_route_reachability` flags a route when its **last static segment** appears
+nowhere in the web source. The file's header already knew literal matching is unsound against a
+client that builds URLs from template literals — that is *why* the rule is the last segment. **It
+assumes the last segment is written out.** Eight routes are built with that segment templated too, so
+their leaf appears nowhere, and every one was frozen in `KNOWN_UNCALLED` as deliberately callerless:
+
+| route | built at | value passed at |
+|---|---|---|
+| `/exports/{cobie,qto,schedule,spaces}.xlsx` | `viewer/tools/exportsSection.ts:57` | the four-entry list on line 54 |
+| `/schedule/{gantt,lob}.svg` | `api/schedule.ts:193-4` | `portal/panels/schedule.ts:731` |
+| `/model/export.{jsonld,parquet}` | `api/model.ts:185` | `portal/panels/standards.ts:488-489` |
+
+The list's own contract is that *"a frozen route that gains a caller must leave"*. These never lacked
+one — eight entries in the allowlist that records deliberate decisions were false, and four of them
+sat under a heading reading *"a user reaches these by clicking a link the server builds, so no client
+caller is expected"*, which is what made them look considered.
+
+## Two mechanisms, because one shape is decidable and the other is not
+
+* **Templated stem** (`exports/${file}.xlsx`) is a **pattern**: the parent segment and the extension
+  must still match literally. Six routes, and dropping the parent anchor vouches for six more that
+  nothing calls.
+* **Templated extension** (`model/export.${fmt}`) is a **two-entry named list with its call site
+  recorded**, because as a pattern it is wrong in both directions: it vouches for six `export.*`
+  routes of which only two are called. The other four are frozen correctly, and each says the same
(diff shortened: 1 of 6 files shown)


ibuilder:
**Blocked on a release tag, not on this diff.**

`API test gate` is red on `94813662`. **658/659 suites pass**, and the one failure is `test_release_current`:

```
shipped v0.3.1132 · newest tag v0.3.1121 · 11 ahead (bound 10)
FAIL  the shipped version is within 10 releases of the newest tag
      v0.3.1132 is bumped but v0.3.1121 is the newest TAG — 11 versions have been merged
      and never released. The auto-updater serves the latest release, so none of it has
      reached a user. Tag the release rather than raising this bound
```

My local run reproduces it identically (658/659, same assertion), so it is not a CI artifact and not a flake.

**Why I have not fixed it here.** The condition is repo-level release hygiene, not anything in this diff — eleven merged versions are untagged and this bump crossed the bound. The gate names its own remedy, *"tag the release rather than raising this bound"*, and I agree with it: raising the bound would be switching off the check that noticed. But pushing a `v*` tag triggers `.github/workflows/desktop.yml`, which creates a GitHub Release and uploads the desktop binaries **the auto-updater serves to users**. That is a publish, so it is the repository owner's call rather than mine, and I have raised it with them rather than doing it.

Nothing else is outstanding: all three CodeRabbit findings are addressed in `94813662` (including a real one — the count was `81 → 75` and is `81 → 73`), and its nitpick turned out to be a performance regression I had introduced, fixed in `1110b70f`: `strip_comments(BLOB)` sat inside a set comprehension and re-stripped a 3.9 MB blob 943 times, making this the slowest gate in the suite. **70.4s → 8.9s**, behaviour unchanged and the new pins still fail under mutation.

This PR stays watched until it is green or closed.

---

PR #391, opened 2026-09-01
Asset-rights NFT stack: mock + EVM mint registry for sealed .mass releases

ibuilder:
## Summary

Completes the asset-rights / NFT stack for sealed `.mass` releases — steps 4 **and** 5.

Each sealed release (`asset_rights.json` + `content_hash`) can now be registered as a mock or on-chain ERC-721 token, with optional IPFS metadata pinning.

## Step 4 (previous commit)

- `ChainProvider` protocol + `MockChainProvider`
- `release_tokens` Postgres table (keyed on `content_hash`)
- `POST /asset-rights/mint`, lookup, project list routes

### Smart contract (`services/chain/`)

- **`MassRelease721`** — OpenZeppelin ERC-721, one token per `content_hash`, owner-only mint
- Foundry tests (5 passing, including fuzz)
- Deploy script: `MASS_RELEASE721_OWNER=0x… forge script script/DeployMassRelease721.s.sol --broadcast`
- CI job runs `forge install` + `forge test`

### Python providers

| Module | Role |
|---|---|
| `evm_provider.py` | Real minting via `web3.py` + JSON-RPC |
| `ipfs_storage.py` | Mock (deterministic CID) or Pinata pinning |
| `release_nft_metadata.py` | ERC-721 JSON from manifest |

### Hardening

- Strict `sha256:<64 hex>` validation on all mint paths
- Auto-pin metadata when `pin_metadata=true` (default)
- Idempotent mint + `IntegrityError` race handling
- EVM gas estimation + EIP-1559 fee fields
- `GET /asset-rights/mints/verify` — registry vs on-chain cross-check
- `/asset-rights/status` now reports `chain.ipfs` and `chain.evm` readiness
- 42 checks in `test_asset_chain.py`

## Configuration

**Mock (default — no network):**
```bash
export AEC_CHAIN_ENABLED=true
export AEC_CHAIN_PROVIDER=mock   # default
```

**EVM (testnet/mainnet):**
```bash
export AEC_CHAIN_ENABLED=true
export AEC_CHAIN_PROVIDER=evm
export AEC_CHAIN_RPC_URL=https://…
export AEC_CHAIN_ID=11155111
export AEC_CHAIN_CONTRACT=0x…
export AEC_CHAIN_MINT_KEY=0x…          # issuer key — never in .mass
export AEC_CHAIN_DEFAULT_RECIPIENT=0x… # optional
```

**IPFS (optional — auto-pin on mint):**
```bash
export AEC_IPFS_PROVIDER=pinata
export AEC_IPFS_PINATA_JWT=…
```

## Test plan

- [x] `python3 test_asset_chain.py` (42 checks)
- [x] `forge test` in `services/chain/`
- [x] `test_alembic_single_head.py`
- [x] ruff on changed modules

<div><a href="https://cursor.com/agents/bc-01a04456-76ba-783e-8c99-1626bca7f622?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-01a04456-76ba-783e-8c99-1626bca7f622&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


ibuilder:
Picked this up with the owner's approval of `web3>=7.0` and the `services/chain` stack. Merged `main` (33 commits, v0.3.1133 → v0.3.1143, **zero conflicts**) and fixed three of the four failures in `117ef6f2`. That merge alone fixed two of them — 656/660 → 658/660 — so part of the red was staleness, not this diff.

| was failing | cause | state |
|---|---|---|
| `pip-compile --generate-hashes` | **provenance, not dependencies** | fixed |
| `alembic upgrade + check` | **a real model/schema divergence** | fixed |
| `API test gate` (route reachability) | two new unreachable routes | fixed |
| `API test gate` (`test_actions_pinned`) | unpinned third-party action | ⛔ **needs you** |

**The lockfile was never wrong about dependencies.** It was generated as `--output-file=requirements.lock.new requirements.in` from inside `services/api/`, so the header recorded the wrong command and all eleven annotations read `# via -r requirements.in` rather than `# via -r services/api/requirements.in`. Regenerated with CI's exact command on Python 3.12 — 84 lines changed and **zero of them a version or a hash**, measured rather than asserted (`diff | grep -E '^[<>] [a-z0-9_.-]+=='` is empty).

**The alembic failure was a genuine defect and the gate was right.** `models.py` declares `content_hash: mapped_column(String, unique=True, index=True)` — one unique index. The migration wrote a named `UniqueConstraint` **and a non-unique index**, so the database carried a constraint the model doesn't declare and lacked the uniqueness it does. `alembic check` reported three pending operations. The migration now matches the model; the guarantee is identical, but the schema stops having drifted from its mapping.

**On the two frozen routes — the part worth reading.** `/asset-rights/mints` and `/projects/{pid}/asset-rights/mints` are frozen with an expiry condition, because this PR touches **no `apps/web` file at all**, so "no client caller" is the honest state of a staged feature. But **four routes were added and the rule can only see two**: `POST /asset-rights/mint` (leaf `mint`) and `GET /asset-rights/mints/verify` (leaf `verify` — the ASSET-VERIFY blind spot, met a second time) are *equally* uncalled and simply invisible to the matcher. Freezing two does not make the other two reachable, and that is written into the entry so the next reader isn't misled.

## ⛔ What I can't do from here

`test_actions_pinned` fails on `foundry-rs/foundry-toolchain@v1` in `ci.yml`. That gate requires third-party actions to be pinned to a 40-hex commit SHA, because a tag is mutable and a workflow step runs upstream code against this repo. `actions/*` and `github/codeql-action` are exempt as first-party; `foundry-rs` is not.

Resolving the real SHA needs read access to `foundry-rs/foundry-toolchain`, which this session isn't scoped for. **I will not invent one** — a fabricated 40-hex value satisfies the regex while pointing at nothing, which is the exact "a check that cannot fail" failure this repo's gates exist to catch. Raised with the owner rather than worked around.

Everything else verified: backend suite **659/660** on this tree (the remaining one being the pin above), `ruff` clean, and `test_lock_satisfies_requirements`, `test_declared_imports`, `test_alembic_single_head`, `test_alembic_migrations` all green.

---

ibuilder:
**Unblocking the one thing this PR was stuck on: the `foundry-rs/foundry-toolchain` SHA.**

This PR adds `- uses: foundry-rs/foundry-toolchain@v1`, an unpinned tag, while every other action in this repository is pinned by commit SHA with a trailing version comment. That pin could not be supplied earlier because the upstream repository was not reachable from the session working this, and inventing a SHA is not an option.

It is reachable now. Resolved by `git ls-remote` against upstream, not by recall:

```
$ git ls-remote https://github.com/foundry-rs/foundry-toolchain refs/tags/v1
908c540300062bd5a7e473851cdb4282204cee09	refs/tags/v1
```

**`refs/tags/v1` → `908c540300062bd5a7e473851cdb4282204cee09`**, which is byte-identical to `refs/tags/v1.9.1` in the same listing — a useful consistency check, since a floating major tag that did *not* match its newest point release would mean the tag had drifted from what it claims to track.

So the line becomes, matching the convention used by `setup-buildx-action`, `build-push-action` and the rest:

```yaml
      - uses: foundry-rs/foundry-toolchain@908c540300062bd5a7e473851cdb4282204cee09  # v1.9.1
```

Two things worth flagging rather than assuming:

- **Pin the version comment to `v1.9.1`, not `v1`.** The comment documents what the SHA *is*; writing `# v1` would record a moving target next to an immovable one, and the next reader could not tell whether the pin had fallen behind.
- **`forge install ... --no-commit`** appears twice in the added job. Recent Foundry removed that flag — `forge install` no longer commits by default, and passing it errors on newer toolchains. Worth confirming against whatever version `v1.9.1` installs before relying on that step, since a green pin with a red install is no better off than it is today.

I have not pushed this: the change belongs on this PR's own branch, which is not mine to write to. Supplying the verified value so it can be applied by whoever picks the PR up.

---

PR #398, opened 2026-09-03
The dependency scan's blocking gates could not block a merge, and a stale comment is why

ibuilder:
# What & why

`security.yml` hosts two **blocking** gates — `scripts/audit_lock_gate.py` and `scripts/audit_npm_gate.py`, each failing the build on an advisory that *has* a published fix — and triggered on push-to-`main`, a weekly schedule, and manifest paths. **Never on `pull_request`.** So neither gate could refuse a merge; each could only report one that had already happened.

On 2026-09-03 that let **four merges (#393–#396) land over a red scan** carrying three pypdf CVEs and two HIGH browserslist advisories, every one already fixed upstream. #397 fixed the advisories. **This fixes the reason they got in.**

Same shape as R39-CONTAINER-PR, whose test file states it as *"a skipped job is not a passed job"* — one level up. There a **job** couldn't run on the event that would show a human the break first; here the whole **workflow** couldn't.

## The stale comment is the finding, not a tidy-up alongside it

The header read:

> Report-only supply-chain scan (pip-audit + npm audit). **Non-blocking by design** — it surfaces known CVEs in the job summary without failing the merge gate (ci.yml stays the hard gate).

True when written, and **false from the moment the blocking gates were added** — and very likely *why* nobody wired the trigger. A reader who believed the header had no reason to run an advisory-only scan on PRs.

This repo already knows that unverified prose drifts. What it hadn't recorded is prose drifting into **the reason a gate was left unable to fire**. Corrected, keeping the old sentence so the record survives.

## Scope of the trigger

Deliberately the **same globs as `push:`**, not left open:

- a PR touching a manifest is the one that can introduce an advisory → audited;
- a PR touching none cannot → not slowed.

Cost **measured** from run 1206 rather than guessed: python 99s, node 21s, sbom 12s, parallel, **~100s wall**. An advisory published upstream against an unchanged manifest is still covered by the schedule and by `push:` — this narrows nothing.

## Three new assertions, each proven by mutation

| mutation | result |
|---|---|
| delete the `pull_request` trigger | fails, naming the consequence |
| diverge one glob between the two lists | fails, printing the asymmetric difference |
| drop a glob from **both** lists | fails on the per-event check alone |

**The third earned its place the hard way.** The first two mutations both tripped the *equality* assertion, which runs first and shadows the per-event one — so the per-event check was not demonstrated by either, and I had no evidence it did anything at all. It needed the case equality cannot catch: both lists identical and both wrong. That division of labour is now in the docstring, marked as established by mutation rather than by design, because it was.

GitHub Actions does not honour YAML anchors, so the duplicated glob list is unavoidable; the equality assertion is the ratchet on it. Neither list is written in the test — both are derived.

## Two pre-existing defects in the same function

- **The paths check was sliced out of the text** as `src[index("    paths:") : index("  workflow_dispatch:")]`. The new `pull_request:` block sits between those bounds, so it would have silently measured a **merged list of twelve globs** and still passed. Parsed as YAML now, per this file's own lesson one function up: *a gate that greps prose measures the wrong text.* The `on:` key is read as both `"on"` and boolean `True`, since YAML 1.1 gives the bare key the latter.
- **`g.lstrip("**/")` strips a character *set*, not a prefix** (ruff B005). Replaced with `removeprefix` after verifying both produce identical output for all six real globs. It surfaced only because ruff was pointed at the file by hand: CI lints `src/ ../data/src/`, which **excludes the api-root test files** — a smaller instance of the same scope-gap family, raised rather than fixed here.

## Verification

- Three mutations fail and the workflow restores byte-identical.
- `ruff` clean on `src/`, `../data/src/`, and on the test file.
- **Full backend suite 659/659** (1101s), including `test_lock_advisories`, `test_npm_advisories`, `test_actions_pinned`, `test_file_sizes`.

- [x] Backend: `ruff check src/ ../data/src/` clean
- [x] Backend: full suite 659/659; the changed test is registered in `run_tests.py` already
- [x] Web: untouched
- [x] No import cycles introduced (no source files changed)
- [ ] `CHANGELOG.md` entry — **not possible without a version bump**; see below
- [ ] Version bumped in both manifests — **deliberately not**; see below
- [x] No secrets, no competitor names

## No version bump

`main` is **v0.3.1143** with **v0.3.1133** the newest tag: 10 ahead of a bound of 10. Any bump re-breaks `test_release_current`. **v0.3.1143 still needs tagging** — that is the third change in a row to say so.

---

---

## Summary by CodeRabbit

* **New Features**
  * Security scans now also run for pull requests that modify dependency manifests or workflow configuration.
  * Workflow documentation now distinguishes report-only scans from blocking security gates and explains actionable advisory behavior.

* **Tests**
  * Added validation for tracked npm manifests, lock files, workflow path filters, and audited lock-file coverage across pushes and pull requests.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -1,8 +1,19 @@
 name: Dependency scan
 
-# Report-only supply-chain scan (pip-audit + npm audit). Non-blocking by design — it surfaces
-# known CVEs in the job summary without failing the merge gate (ci.yml stays the hard gate).
-# Runs weekly and on dependency-manifest changes.
+# Supply-chain scan (pip-audit + npm audit + bandit + gitleaks). Each tool appears TWICE: a
+# report-only step that surfaces everything into the job summary, and a BLOCKING gate over the
+# subset that is unambiguously actionable. `scripts/audit_lock_gate.py` and
+# `scripts/audit_npm_gate.py` fail the build on an advisory that HAS a published fix.
+#
+# THIS HEADER USED TO SAY "Non-blocking by design ... ci.yml stays the hard gate", and it was
+# false from the moment those gates were added — which is very likely WHY this workflow had no
+# `pull_request:` trigger for as long as it didn't. A reader who believed the header had no reason
+# to wire an advisory-only scan into PRs. On 2026-09-03 that gap let FOUR merges (#393-#396) land
+# over a red scan carrying three pypdf CVEs and two HIGH browserslist advisories, every one of them
+# already fixed upstream. A stale comment is not a documentation problem when it is the reason a
+# gate was left unable to fire.
+#
+# Runs on dependency-manifest changes (push AND pull_request — see below), plus weekly.
 
 on:
   schedule:
@@ -22,6 +33,30 @@ on:
       - "**/package.json"
       - "**/package-lock.json"
       - ".github/workflows/security.yml"
+
+  # THE BLOCKING GATES ABOVE COULD NOT BLOCK A MERGE UNTIL THIS EXISTED. With only `push:`, they
+  # fire AFTER a PR lands — they report a merge that already happened. Same shape as
+  # R39-CONTAINER-PR ("a skipped job is not a passed job"), one level up: there a JOB could not run
+  # on the event that would let a human see the break first; here the whole WORKFLOW could not.
+  #
+  # Deliberately scoped to the SAME globs as `push:`, not left open:
+  #   * a PR that touches a manifest is the one that can introduce an advisory, and it gets audited;
+  #   * a PR that touches no manifest cannot, and is not slowed by ~2 minutes of scanning.
+  # An advisory published upstream against an UNCHANGED manifest is still caught by the weekly
+  # schedule and by `push:` — this trigger narrows nothing that was previously covered.
+  #
+  # The two glob lists must stay identical, and GitHub Actions does not honour YAML anchors, so the
+  # duplication is real and `test_lock_advisories.py` asserts the two lists are equal rather than
+  # trusting either to remember the other.
+  pull_request:
+    paths:
+      - "**/requirements*.txt"
+      - "**/requirements*.lock"
+      - "**/requirements*.in"
+      - "**/package.json"
+      - "**/package-lock.json"
+      - ".github/workflows/security.yml"
+
   workflow_dispatch:
 
 permissions:
--- a/services/api/test_lock_advisories.py
+++ b/services/api/test_lock_advisories.py
@@ -25,7 +25,8 @@ import sys
 from datetime import date
 
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", ".."))
-_GATE = os.path.join(os.path.dirname(__file__), "..", "..", "scripts", "audit_lock_gate.py")
+ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "..")
+_GATE = os.path.join(ROOT, "scripts", "audit_lock_gate.py")
 
 
 def _load():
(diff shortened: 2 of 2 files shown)


ibuilder:
## The fix verified itself on this PR

The strongest evidence here isn't an assertion about the workflow — it's the workflow running on an event it structurally could not run on before this commit:

```
name:  "Dependency scan"
event: "pull_request"        ← has never happened on any PR
run:   1207                  (1206 was the last push-triggered run)
path:  .github/workflows/security.yml
pull_requests: [398]
```

All three of its jobs are in this PR's check list — `python`, `node`, `sbom` — and all three passed. Step-level, on the `python` job, that means these ran **on a pull request** for the first time:

- `pip-audit — GATE on advisories that HAVE a published fix (blocking)` ✅
- `bandit — GATE on HIGH severity (blocking)` ✅
- `gitleaks — secret scan of the full git history` ✅
- `MCP tool-poisoning self-audit` ✅

and on `node`:

- `npm audit — GATE on advisories that HAVE a published fix (blocking)` ✅

Added cost to this PR: ~20s (`node`), ~12s (`sbom`), ~106s (`python`), all in parallel with a 23-minute API gate. Not on the critical path.

## One thing this does not fix, stated plainly

This closes the hole for a PR that **touches a manifest**. It does nothing for an advisory published upstream against an *unchanged* manifest — that case was, and remains, covered only by the weekly schedule and by `push:`. The trigger narrows nothing previously covered, but it should not be read as making the gates comprehensive. The window between a Monday scan and a Wednesday advisory is unchanged.

## Also raised, deliberately not fixed here

CI lints `src/ ../data/src/`, which **excludes the api-root test files**. That is the only reason the pre-existing `lstrip("**/")` (ruff B005) in this very function went unnoticed — I found it by pointing ruff at the file by hand, not because anything failed. Same scope-gap family as the defect this PR fixes, one layer over. Worth its own change rather than being folded into this one.

---

ibuilder:
@coderabbitai the moderate-risk finding was correct, and the gap was in the check I had just written. Fixed in `6b3cdd30`.

**What you found.** `test_every_audited_path_can_actually_trigger_the_workflow` iterates `audit_lock_gate.LOCKS`, which holds exactly one entry — `services/api/requirements.lock`. Remove `**/package.json` *and* `**/package-lock.json` from **both** trigger lists and every assertion in the file stays green: the two lists still match each other, and the single audited path they know about still matches a surviving glob. The blocking npm gate stops firing on manifest changes, silently, with the tests reporting the workflow fully covered.

**Why I missed it, which is the part worth recording.** I verified that check by mutation and concluded it was sound. It *was* sound — for the half it derived. Its population was one Python lockfile while its name claimed the workflow's triggers generally. That is the same failure this PR fixes, one layer in: **a coverage check is only as wide as the population behind it.** My mutation exercised the Python side and I generalised from it.

**The fix.** `audit_npm_gate.py` declares no path list to compare against — it shells out to `npm audit`, whose subject is implicit — so the population comes from `git ls-files` rather than a constant:

```
package.json
package-lock.json
apps/web/package.json
services/converter/package.json
```

Four today; adding a workspace tomorrow puts it in scope automatically, which a hardcoded list would not. Same side this repo closed the container-matrix gap from, for the same reason. It carries its own vacuity guard — an empty population passes every assertion while measuring nothing, and `git ls-files` returns empty outside a repo — so it requires a non-empty result *and* a `package-lock.json` in it, that being the file `npm audit` actually reads.

Both coverage checks now share one `_matches_a_glob` helper. Two matchers for one question drift into disagreeing about what "covered" means.

**Verified by your exact scenario**, not by reading it back — dropping both npm globs from both lists:

```
push == pull_request? True | globs: 4
PASS  push and pull_request audit the same manifests   4 glob(s), identical
AssertionError: these tracked npm manifests match none of security.yml's `push` globs
  [...]: ['apps/web/package.json', 'package-lock.json', 'package.json',
  'services/converter/package.json'] — changing one would not run the blocking npm gate that audits it
```

Note the equality assertion still **passes** there at 4 identical globs — which is the confirmation that the gap was genuinely invisible before, rather than merely unlikely. Restores clean; ruff clean.

---

PR #403, opened 2026-09-03
SCALE-SEAM (87) + RUFF-SCOPE — the map described the tail, and the lint described two directories

ibuilder:
Two commits, one theme: **a check that was real, beside a sentence about its reach that was not.**

---

# 1. SCALE-SEAM (87) — the map described the tail, not the file

Thirteen methods out of `client.ts` (**1,131 → 1,015**) into eight existing mixins.

| destination | methods | what they answer |
|---|---|---|
| `elements.ts` | `materialPalette`, `saveMaterialPalette`, `applyMaterialPalette` | what does the model look like? |
| `entitlements.ts` | `testFitCompare`, `testFitOptimize` | does a scheme fit the site? (beside `feasibility`) |
| `proforma.ts` | `property`, `saveProperty` | what is the asset's record? |
| `estimate.ts` | `rebarBbs`, `rebarBbsCsvUrl` | what does it cost to build? |
| `modules.ts` | `complianceExpiring` | what is expiring? |
| `procurement.ts` | `bidLeveling` | how did the bids compare? |
| `schedule.ts` | `safetyMetrics` | is the job safe? |
| `evm.ts` | `pxSummary` | how is the programme performing? |

`OptScheme`, `DepthPoint`, `MaterialEntry`, `MaterialPaletteResult`, `EgressResult` moved to `types.ts` with them.

### The finding

Deriving the whole population instead of trusting the banner: **130 methods, 4 below it, 126 above it.** Those 126 — `disciplineTree`, `classify`, `specManual`, `editUndo`, `energyModel`, `rentRollScrub`, `esgSummary` and the rest — were never inside the RACI or CX-1 banners, so no map has ever covered them. The UNFILED map described the **tail** of the file, and every "N remain" made against it was scoped to a list the same slices had authored.

Banner renamed **UNFILED → STAYING**, reduced to the four global methods it actually decides (`enumOptions`, `searchAll`, `attachmentUrl`, `templates`), and it now states in its own text that this is not the end of SCALE-SEAM.

### Correction: the predicate was replaced twice, and I reported the first replacement as a fix

`test_roadmap_status.py` held SCALE-SEAM open on `client.ts > 1200` — a threshold proxy, which decayed and went red at (85). I replaced it with a check for the banner string. **That replacement was also a proxy**: it would have declared SCALE-SEAM DONE with 126 methods outstanding, and I described it as a fix in a commit message, in #402's body, and to the user before noticing.

It now counts methods not on the declared keep-list, mutation-checked four ways: baseline 126 → OPEN · banner renamed → still OPEN · keep-list deleted → 130, fails safe · simulated completion → 0 → DONE.

### Two corrections to earlier slices

`pxSummary` is the programme-level rollup that belonged with the rollup/detail split `evm.ts` made at ⓽ and didn't move with it. `modules.ts` carried a doc comment orphaned from `complianceExpiring` — swept in without its method. Both reunited.

### The ratchet now checks its own history chain

`test_file_sizes.py` records each slice as `(before -> after)` beside the pin; at (86) CodeRabbit caught `(1173 -> 1132)` against a pin of `1131`. Consecutive pairs must now link; an unrecorded **shrink** fails, growth does not. Four tries, all four failures found by running it: `\d+` skipped underscores, a closing-paren requirement invented a false break, strict equality flagged legitimate growth, and the finished checks **printed FAIL but exited 0** because they sat below the `if FAILED:` block. `viewer/app.ts` carries a named exemption for its documented 2,571/2,570 slack — calling that a defect would have meant falsifying a slice's measurement to satisfy a gate.

---

# 2. RUFF-SCOPE — the lint was real; the scope was the fiction

`ci.yml` ran `cd services/api && ruff check src/ ../data/src/`. That is two source trees. The repository tracks **1,338 `.py` files**; those two hold 612.

**The other 726 were never linted at all:**

| | |
|---:|---|
| 672 | `services/api/*.py` — the entire backend test suite |
| 27 | `services/api/migrations/versions/` — every alembic revision |
| 7 | `scripts/` |
| 4 | `integrations/pyrevit/` — the Revit bridge |
| 3 | `apps/editor-bridge/` — the Blender bridge |
| 3 | `services/data/*.py` |
| … | `plugins/`, `.claude/hooks/` |

The step exited 0 and printed *"All checks passed!"* on every push. A green lint step reads as "this repository is linted" — the check was real, the **scope** was the fiction.

Now `cd services/api && ruff check ../..` — one path, no directory list to fall out of date. **The `cd` is load-bearing:** `services/api/ruff.toml` is the only ruff config in the tree, and ruff resolves it *and its `extend-exclude` paths* from the working directory. Running from the root with `--config services/api/ruff.toml` finds the config but resolves the excludes against the wrong root, silently un-excluding all five vendored trees — measured at 161 findings in code we copy verbatim and must not edit.

**230 findings → 0.** 166 auto-fixed, 40 with `--unsafe-fixes` reviewed hunk by hunk, the last six hand-written. Substantive: **five `assert False`** (B011), one guarding path-traversal rejection in `test_security.py`. Nothing here runs `python -O`, so those were **latent, not live** — but they are exactly what ruff exists to catch, and it had never been asked.

### Widening the scope broke something, and only reading the diff caught it

`--fix` over the new scope stripped `# -*- coding: utf-8 -*-` from four files under `integrations/pyrevit/`, **three of which contain non-ASCII source**. That tree runs inside Revit on pyRevit's **IronPython 2.7** engine — `massing_api.py` says so in its first docstring line and proves it by try/excepting `urllib.request` against `urllib2`. Python 2 defaults to ASCII source encoding, so removing that line is `SyntaxError: Non-ASCII character` **at import**, in code no CI here executes.

`target-version` is one global setting and can't be narrowed per directory, so `ruff.toml` now disables `UP` for that tree with the reason beside it. `F`, `E9`, `B`, `C4` stay on — all 2.7-compatible findings worth catching. **Scope and rule set are different questions, and widening one can break the other.**

### The gate reads the workflow rather than restating it

`services/api/test_ruff_scope.py` asserts every tracked `.py` is either in ruff's file set or under one of five explicitly named vendored prefixes — and it gets that file set by **parsing the `ruff check` line out of `ci.yml`** and running `--show-files` on it. A scope gate that measures a command nobody runs is the same bug one level up. Each vendored prefix must also still match a tracked file, so a renamed tree can't take a carve-out with it.

Mutation-checked: CI narrowed back to `src/ ../data/src/` → **exit 1**, naming the unlinted files; a vendor prefix renamed → **exit 1**, naming the stale entry; restored → **exit 0**. Exit codes read directly, not through a pipe.

### Roadmap correction

The roadmap's *"the cheap dead-code classes are already gated — `ruff --select F401,F841` passes clean across `services/api/src` and `services/data/src`"* bullet **named its scope correctly and I still read a tree-wide conclusion out of it**. Widening found five F401 and one F841 outside that scope. The bullet's argument survives; its reach did not. Both ruff claims in the roadmap are corrected in place.

---

## Verification

`tsc --noEmit` exit 0 · eslint exit 0 · **203 files / 2,059 web tests** · `ruff check ../..` clean over 1,280 files · `test_file_sizes`, `test_roadmap_status`, `test_claude_md_gates`, `test_changelog_current`, `test_reachable`, `test_declared_imports`, `test_ruff_scope` all OK.

The full backend suite is running locally against both commits (0 failures so far, including `test_security` and `test_compute_graph` — the two files whose `assert False` was rewritten). CI's API test gate is the authority and is running on this head; I'll drive any failure here.


ibuilder:
@coderabbitai review

Two things worth aiming at, since the diff is large and mostly mechanical:

1. **`services/api/test_ruff_scope.py`** — it parses the `ruff check` line out of `.github/workflows/ci.yml` and asserts that command's file set covers every tracked `.py` outside five named vendored prefixes. If that regex can be satisfied by a `ci.yml` edit that narrows the real scope, the gate is a proxy and I'd like to know.

2. **The `--unsafe-fixes` hunks.** 40 findings were fixed with `ruff --unsafe-fixes` (B011 `assert False` → `raise AssertionError`, C414/C416/C402/C408, B007). I reviewed each, but they touch ~125 test files and behaviour-preservation is the whole question. `test_security.py:101` and `test_compute_graph.py` are the ones where a wrong rewrite would silently weaken a guard.

Context on the third: the same `--fix` run stripped `# -*- coding: utf-8 -*-` from four `integrations/pyrevit/` files (UP009). That tree runs on IronPython 2.7 inside Revit and three of those files hold non-ASCII source, so it was a `SyntaxError` at import in code no CI here executes. That is reverted and `ruff.toml` now disables `UP` for that path only — but it's the shape of mistake worth a second pair of eyes across the rest of the diff.

---

ibuilder:
@coderabbitai The scope-gate bypass is real and is fixed in `17463454`. Thank you — that is the finding I most wanted and least expected to be told yes on.

**What was wrong:** `re.search` returns the *first* match. Your construction works exactly as described — an earlier `ruff check ../.. --show-files` plus a narrowed enforcing command below it leaves the gate measuring the decoy.

**The fix** parses every workflow under `.github/workflows/` with PyYAML, collects **every** `run:` step mentioning `ruff check`, and requires **exactly one**. I went with the stricter of your two suggestions: two invocations is ambiguous, this gate cannot know which one enforces, and *ambiguity resolved by picking one* is precisely the bypass. If a second is ever genuinely wanted, the gate has to be taught which is the lint by step name — it says so in the failure text, so nobody relaxes the count instead.

A step that cannot fail the build is now rejected outright: `--exit-zero`, `--show-files`, `--statistics`, `--show-settings`, a trailing `|| true`, and `continue-on-error` on either the step or its job. That last one needed structural YAML rather than a regex, which is why PyYAML.

Five mutations, all red, restored green: decoy-then-narrow · `--exit-zero` · `|| true` · `continue-on-error` · narrowed single invocation.

**Two follow-ons your finding shook loose:**

1. The new `import yaml` was undeclared in both `requirements.in` and `requirements-dev.txt`, reaching the lock only `# via` fastapi/uvicorn/pyHanko/bandit/starlette/markdown-it-py — and `test_declared_imports.py` stayed green, because it walks `services/api/src` and `services/data/src` only, and every gate file lives in `services/api/` itself. Same scope hole this PR fixes for ruff, one directory over. pyyaml is now declared; the class is recorded in `docs/roadmap.md` as open.

2. `requirements-dev.txt`'s ruff floor comment said the pin had been validated "against the exact scope ci.yml lints, `src/ ../data/src/`" — which this PR makes stale, so the validation behind that pin covered 612 of 1,280 files. Corrected in place.

Confirming your other two results independently: the `assert False` → `raise AssertionError` rewrites preserve behaviour (`AssertionError` isn't caught by the surrounding `except ValueError:`, so the guards still fail when the expected exception doesn't occur), and the pyRevit encoding declarations are intact — that revert is why `ruff.toml` now carries the `UP` per-file-ignore.

---

PR #404, opened 2026-09-03
SCALE-SEAM (88) — the operate-phase cluster, first slice out of the unmapped 126

ibuilder:
Nine methods out of `client.ts` (**1,015 → 953**) into a new `apps/web/src/api/operations.ts` — the first slice taken from the 126 methods (87) found had never been in any map.

## What they answer together: the building is built and running

| | methods | question |
|---|---|---|
| maintained? | `cmmsGeneratePm`, `cmmsKpis` | PM work orders, compliance, MTTR |
| consuming what? | `energyActual`, `energyBenchmarkStatus`, `esgSummary` | metered EUI, GHG from actuals, water, POE actual-vs-design |
| what condition? | `fcaIndex`, `fcaPortfolio`, `reserveStudy` | FCI, deferred-maintenance backlog, whether the reserve fund clears the replacement schedule |
| operable at all? | `twinReadiness` | is the asset data good enough — the precondition for the other three |

**Checked against the backend rather than assumed.** All nine are served by `services/api/src/aec_api/routers/operations.py`, whose own docstring names the same cluster. That is corroboration from a source this extraction did not write — which is the whole point after (87), where a completeness claim computed over a self-authored list turned out to be confident and unfounded.

## Three did not come, and the words say otherwise on every one

**`lifecycle` / `lifecycleSeed` are not asset lifecycle.** Their banner reads *"design lifecycle (RIBA/AIA phases + itemized soft costs)"* and the payload is `design_fee_pct`, `deliverables`, `iso_status`, `soft_costs` — design stage gates and what the design costs, years before anything operates. **Fourth shared-word trap in this sequence**, after entitlements (land-use vs licence), view (grid filter vs 3D selection), and now carbon.

**`projectCarbon` is EMBODIED carbon** — what building it emits, a design-phase estimate. The GHG in `esgSummary` comes from metered utility data. Same molecule, opposite ends of the asset's life.

**`camReconciliation` shared `reserveStudy`'s banner *and* its backend router, and still stayed.** It reconciles budget against actual operating expense, then allocates the recoverable pool **across tenants** by share and returns `balance_due` per suite — a lease-revenue answer, not a building-condition one. It waits for a rent-roll slice with `rentRollScrub`, `netEffectiveRent` and `normalizeT12`. Moving it along to empty a banner would have been exactly the route-and-neighbour reasoning this sequence keeps rejecting, so the banner it sat under is **rewritten** rather than left naming a method that has gone.

## (87)'s own banner had already drifted

The STAYING banner listed `esgSummary` among its examples of work still above the line. This slice moved it — so a banner offering evidence of what is *unfinished* was naming something finished, about two hours after it was written.

Corrected to **117**, re-pointed at `camReconciliation`, and it now instructs the reader to **re-derive the number and the names from the file rather than read them there**. The count itself is asserted by `services/api/test_roadmap_status.py`, which counts methods absent from the declared keep-list.

The instance is small; the class is the one this whole sequence is about. Every example list in a comment is a claim with a shelf life.

## Still open, stated plainly

**117 methods remain above the banner and there is still no map for them.** This was one cluster read out of the file, not a plan for the rest. `test_roadmap_status` keeps SCALE-SEAM OPEN and will until the file is actually finished.

## Verification

`tsc --noEmit` exit 0 · eslint exit 0 · **203 files / 2,059 web tests** · `ruff check ../..` clean over the whole tree · `test_file_sizes`, `test_roadmap_status`, `test_ruff_scope`, `test_claude_md_gates`, `test_changelog_current`, `test_reachable`, `test_declared_imports` all OK.

The history-chain gate added in (87) validated its own new `(1015 -> 953)` entry — first time it has checked a link it did not also write.

---

## Summary by CodeRabbit

- **Refactor**
  - Organized operations-related API capabilities into a dedicated area while preserving existing access patterns.
  - Continued support for maintenance, energy, facility condition, reserve study, ESG, and digital twin readiness operations.

- **Documentation**
  - Updated the changelog and roadmap to reflect the latest API organization and method counts.
  - Corrected progress-tracking details and examples for the ongoing API restructuring.


ibuilder:
@coderabbitai review

This one is ~6 files, well under the limit that skipped #403. Three things worth aiming at:

1. **The mixin wiring in `client.ts`.** `withOperations(...)` was inserted into a 37-deep nesting of mixin factories by string surgery, and the closing paren was added by matching `(HttpCore)`. `tsc` is clean and the 2,059 web tests pass, but a mis-nesting that still typechecks is exactly the kind of thing worth a second look.

2. **Did anything get left behind by the extraction?** The nine methods were cut with a brace-counting script that also walks back over contiguous `//` comment lines. I know it correctly carried two banners across and correctly stranded a third (`camReconciliation` lost the banner it shared with `reserveStudy`, and I rewrote it). What I'd like checked is the inverse: a moved method that lost its comment, or a type still imported in `client.ts` with no remaining reader.

3. **The judgment call.** `camReconciliation` sits in the same backend router as `reserveStudy` and shared its banner, and I still held it back — it allocates recoverable opex across tenants and returns `balance_due` per suite, which reads as lease revenue rather than building condition. If you think co-location in `operations.py` should have won, say so; I'd rather argue it now than discover it two slices later.

Context: on #403 you found a real bypass in `test_ruff_scope.py` (`re.search` taking the first match). That is fixed and merged.

---

PR #405, opened 2026-09-03
DECLARED-SCOPE — the same hole one directory over, hiding a real one

ibuilder:
`test_declared_imports.py` exists because `httpx` was declared in no requirements file and reached the lock only `# via anthropic` — so bumping that SDK would have deleted a package two shipped modules import, with nothing going red until two routes 500'd.

It walked `services/api/src` and `services/data/src`. **Every one of the ~30 gate files lives in `services/api/` itself, outside that walk.** That is RUFF-SCOPE's defect exactly, in the checker written to catch this class: the check was real, its reach was the fiction. Proven on #403, where that PR's own `import yaml` went undeclared and this gate stayed green.

## Widening it found one true instance immediately

`test_saml.py` does `from lxml import etree` at module scope.

- `lxml` is declared in **neither** `requirements.in` nor `requirements-dev.txt`.
- It arrives in `requirements.lock` only `# via` pyHanko / signxml / python-docx.
- signxml's own pin is `lxml<7,>=5.2.1` — so a signxml release that changed XML libraries would remove a package this suite imports **directly**, and nothing would go red until the SAML tests hit an `ImportError`.

Undeclared for as long as `test_saml.py` has existed. Now in `requirements-dev.txt` with its provenance written beside it.

## Two scopes, because a dev pin does not ship

| scope | may declare in | why |
|---|---|---|
| gate files | `requirements.in` **or** `requirements-dev.txt` | test-only code never enters the image |
| shipped modules | `requirements.in` only | that is what compiles into the runtime image |

A runtime import backed solely by a dev pin passes CI and `ImportError`s in production. Proven rather than asserted: a module-scope `import yaml` added to `src/aec_api/__init__.py` **fails**, while the same import in `test_ruff_scope.py` passes.

## The widening's own first output was five false positives

`run_tests`, `mcp_server`, `vendor_drift`, `massing_api`, `massing_export` all reported as *"not installed"*. None is a package — they are sibling `.py` files, plus the pyRevit bridge library that `test_revit_bridge.py` puts on `sys.path`. The first-party set was derived from `src/` package **directories**, which cannot see a module that is a bare file.

Now enumerated from the importable directories, and scoped to those deliberately: a repository-wide basename sweep would forgive a real third-party import that happened to share a name with one of our files, and forgiving is the failure this test exists to prevent.

**A widened scope's first run is data, not a verdict.** Five of the six findings were the gate's own fault; the sixth had been true for months. Reporting the raw six as "six undeclared dependencies" would have been wrong in the expensive direction — it is the same error as trusting a self-authored list, inverted.

## Verification

Mutation-checked three ways, exit codes read directly:

| mutation | result |
|---|---|
| `lxml` undeclared again | **exit 1** — the real finding returns |
| gate dirs pointed at an empty path | **exit 1** on the vacuity guard |
| dev-only `import yaml` in a shipped module | **exit 1** — the two scopes are genuinely separate |
| restored | **exit 0** |

`test_declared_imports`, `test_ruff_scope`, `test_file_sizes`, `test_roadmap_status`, `test_claude_md_gates`, `test_changelog_current`, `test_reachable` all OK · `ruff check ../..` clean tree-wide.

Backend-only change (one gate file plus a requirements declaration); no frontend surface touched.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dependency checks to cover both shipped application modules and development tooling.
  * Corrected dependency classification for packaged entry points, preventing inaccurate warnings.
  * Declared `lxml` for development and test workflows.

* **Documentation**
  * Updated the roadmap and changelog to reflect the completed dependency-scope improvements and identified dependency requirement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -675,6 +675,47 @@ comment is a claim with a shelf life.*
 Pin 1,015 → 953. **117 above the banner, and still no map for them** — this was one cluster read out of
 the file, not a plan for the rest.
 
+Twenty-seventh follow-on on the same version: **DECLARED-SCOPE** — *the same hole, one directory
+over, and it was hiding a real one*.
+
+`test_declared_imports.py` exists because `httpx` was declared in no requirements file and reached the
+lock only `# via anthropic`, so an SDK bump would have deleted a package two shipped modules import.
+It walked `services/api/src` and `services/data/src` — **and every one of the ~30 gate files lives in
+`services/api/` itself, outside that walk.** RUFF-SCOPE's defect exactly, in the checker written to
+catch this class: *the check was real, its reach was the fiction.*
+
+**Widening it found one true instance immediately.** `test_saml.py` does `from lxml import etree` at
+module scope. `lxml` is declared in neither requirements file and arrives in the lock only `# via`
+pyHanko / signxml / python-docx. signxml's own pin is `lxml<7,>=5.2.1`, so a signxml release that
+changed XML libraries would remove a package this suite imports directly — and nothing would go red
+until the SAML tests hit an ImportError. Undeclared for as long as `test_saml.py` has existed. Now in
+`requirements-dev.txt`, with the provenance written next to it.
+
+### Two scopes, because a dev pin does not ship
+
+A gate may import from `requirements.in` **or** `requirements-dev.txt`; a shipped module may import
+only from `requirements.in`, because that is what compiles into the runtime image. A runtime import
+backed solely by a dev pin passes CI and `ImportError`s in production. Proven rather than asserted: a
+module-scope `import yaml` added to `src/aec_api/__init__.py` **fails** (`yaml` is a dev-only pin),
+while the same import in `test_ruff_scope.py` passes.
+
+### The widening's own first output was five false positives
+
+`run_tests`, `mcp_server`, `vendor_drift`, `massing_api`, `massing_export` all reported as *"not
+installed"*. None is a package: they are sibling `.py` files, plus the pyRevit bridge library that
+`test_revit_bridge.py` puts on `sys.path`. The first-party set was derived from `src/` package
+**directories**, which cannot see a module that is a bare file. Now enumerated from the importable
+directories — scoped to those, deliberately, since a repository-wide basename sweep would forgive a
+real third-party import that happened to share a name with one of our files, and forgiving is the
+failure this test exists to prevent.
+
+*A widened scope's first run is data, not a verdict.* Five of the six findings were the gate's fault;
+the sixth had been true for months.
+
+Mutation-checked: lxml undeclared → exit 1 · gate dirs pointed at an empty path → exit 1 on the
+vacuity guard · a dev-only import in a shipped module → exit 1. Restored → exit 0.
+
+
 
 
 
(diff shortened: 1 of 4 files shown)


ibuilder:
@coderabbitai review

Two files. The interesting question is whether the widened gate can be made to forgive something it should catch — you found exactly that shape of hole in `test_ruff_scope.py` on #403 (`re.search` taking the first match), so please aim there:

1. **`LOCAL_MODULES`** enumerates `.py` stems from three directories and adds them to the first-party set, so a bare-name import of a sibling file isn't reported as a missing package. That means **any third-party package whose import name collides with one of our filenames is silently forgiven.** I scoped it to importable directories rather than sweeping the repo for exactly this reason, but I'd like the collision surface checked against what's actually installed.

2. **The two-scope split.** Gate files may declare in `requirements.in` or `requirements-dev.txt`; shipped modules only in `requirements.in`. I'd like the *partition* checked — is there a path that lands in neither bucket, or in the looser one when it should be in the stricter? `_walk_py(recursive=False)` on the gate dirs is deliberate (migrations/, `_models/`, scripts/ are separate populations), but deliberate isn't the same as correct.

3. **`module_scope_imports`** is pre-existing and I didn't touch it, but this change is the first to point it at test files, which use `sys.path` manipulation and conditional imports far more than `src/` does. If it misclassifies anything in that idiom, this PR is where it starts mattering.

Context on the finding itself: `lxml` was imported at module scope by `test_saml.py` and declared in neither requirements file, arriving only `# via` pyHanko/signxml/python-docx. That is real and now declared. Five other findings from the same run were my gate's own false positives — sibling modules — and are fixed rather than suppressed.

---

ibuilder:
@coderabbitai P1 confirmed and fixed in `39beaa0f`. You were right, and the framing matters: **the split had reintroduced its own defect.** The two-scope partition exists to stop a dev-only pin backing a runtime import, and I then put two shipping files in the dev bucket by assuming "directly under `services/api`" means "test".

Verified both against primary sources rather than taking the names:
- `desktop.spec:49` and `sidecar.spec:43` both pass `["desktop_entry.py"]` as the analysis script.
- `Dockerfile:67` `COPY`s `seed_demo.py` into `/app`, and `docker-compose.yml:135` runs it in the seed profile.

**Fixed by derivation, not by naming them.** The shipped set now comes from the packaging manifests themselves — both `.spec` files, both `Dockerfiles`, `docker-compose.yml` — so any `.py` they name clears the stricter scope wherever it sits on disk. Listing the two would fix today and leave the *next* packaged entry point to land in the loose bucket silently; that's the same "a hand-list goes stale" argument behind `FIRST_PARTY` and `LOCAL_MODULES`. An independent sweep of every direct `.py` against those manifests returns exactly your two, so the derivation and your finding agree.

A vacuity guard came with it: if the manifests ever stop naming any direct `.py`, the gate fails rather than quietly moving every entry point into the looser scope.

Mutations, exit codes read directly:

| | |
|---|---|
| dev-only `import yaml` in `desktop_entry.py` | **exit 1 as `[shipped]`** — it passed as `[gate]` before the fix |
| same import in a genuine test file | **exit 0** |
| `PACKAGING` pointed at files naming no `.py` | **exit 1** on the vacuity guard |

On your other two answers: agreed that `LOCAL_MODULES` suppresses only the five expected siblings, and noted that the import-name collision check was **inconclusive** because the checkout lacks the API venv's distribution metadata — I'd rather that stay recorded as inconclusive than be read as clean. The collision surface is bounded by scoping `LOCAL_MODULES` to three importable directories rather than sweeping the repo, but bounded is not the same as verified.

Three scope defects today — the linter's, this gate's, and this gate's own fix — and the third was introduced while correcting the second. Widening a scope is itself a scoping decision.

---

PR #406, opened 2026-09-03
SCALE-SEAM (89) — two clusters, and a banner that named two of its four

ibuilder:
Eight methods out of `client.ts` (**953 → 906**) into two new mixins. Second slice from the 117 that sit above the STAYING banner.

## `resilience.ts` — what could the environment do to this project?

One question asked of three different things, plus the composite over all three:

| | asked of | |
|---|---|---|
| `resilienceFlood` | the SITE | SFHA, design flood elevation, which assets sit below it |
| `resilienceStormwater` | the CIVIL DESIGN | runoff coefficient, peak CFS, detention volume |
| `resilienceWeather` | the PROGRAMME | weather-sensitive activities by trade, open site risks, delay days |
| `resilienceClimateRisk` | all three | one rating, and the factors behind it |

**The tension is named rather than smoothed over.** `resilienceWeather` reads activities with trade, start, finish and percent complete — that is `schedule.ts` vocabulary, and a split on subject matter would send it there. It stays because the question is not *"how is the programme going"* but *"what is the weather doing to it"*, and because `resilienceClimateRisk` folds its `weather_delay_days` and `high_severity_open` into the same composite as the flood and runoff figures. Separating an input from the rollup that consumes it is the split ⓽ made and (87) had to undo.

## `responsibility.ts` — who is accountable for what?

The backend agrees **exactly**: `aec_api/routers/responsibility.py` holds those four routes and nothing else. A 1:1 router correspondence is the strongest corroboration available for a grouping, and it was checked rather than assumed — the same evidence that settled (88).

**It did not go to `modules.ts`, and the router's own docstring is why it nearly did:**

> Rows themselves are ordinary `responsibility` module records, so create/edit/delete of individual cells goes through the generic /modules CRUD.

Same storage, and that CRUD lives in `modules.ts`. **Storage is a HOW.** A caller reaching for `modules.ts` wants records of a module type; a caller reaching for these wants to know who is *Accountable* for a task and to seed a grid that has none. "They are all module records" is (85)'s "they are all multipart uploads" one layer down.

Four methods is a small file, and that is fine — `risk.ts`, `dealMemory.ts` and `assetRights.ts` hold two each; `designOptions.ts` and `library.ts` three. Sizing a mixin to feel substantial is not a reason to file a method where it does not answer.

## The banner over-claimed, for the sixth time in this sequence

`// --- climate & water resilience (flood + stormwater) ---` sat above **four** methods and named **two**. Weather and the climate-risk composite were filed at whatever header happened to be nearest — the same defect as ⓽'s "AI drafting" running on into sheet extraction, (81)'s `ifcClassify` under a G704 turnover header, and (82)'s RACI banner covering thirteen methods while describing four.

`ResponsibilityMatrix` left `client.ts`'s type imports with the methods; it had no other reader.

## Still open, stated plainly

**109 methods remain above the STAYING banner and there is still no map for them.** The banner's seven example names were checked against the file this time and none had moved — *that check is the point, not the result*; it drifted at (88) by naming a method that slice had just moved.

## Verification

`tsc --noEmit` exit 0 · eslint exit 0 · **203 files / 2,059 web tests** · `ruff check ../..` clean tree-wide · `test_file_sizes`, `test_roadmap_status`, `test_ruff_scope`, `test_claude_md_gates`, `test_changelog_current`, `test_reachable`, `test_declared_imports` all OK.

The ratchet's history chain validated its own new `(953 -> 906)` entry.

---

## Summary by CodeRabbit

- **Refactor**
  - Environmental resilience and responsibility API operations are now organized into dedicated service areas.
  - Existing capabilities for flood, stormwater, weather, climate risk, responsibility matrices, templates, and configuration remain available without functional changes.

- **Documentation**
  - Updated the roadmap and changelog to reflect the API organization changes and current implementation progress.


ibuilder:
@coderabbitai review

You've found a real defect on each of the last two PRs — the `re.search` first-match bypass on #403 and the packaged-entry-point partition on #405. Both were places where I'd reasoned my way to a conclusion and the reasoning had a hole. So please aim at the reasoning here, not just the mechanics:

1. **Is `resilienceWeather` in the right mixin?** It returns `weather_sensitive_activities` with `trade`, `start`, `finish`, `percent` and `delay_reports` — that is `schedule.ts`'s vocabulary, and a split on subject matter sends it there. My argument for keeping it in `resilience.ts` is that `resilienceClimateRisk` consumes its `weather_delay_days` and `high_severity_open` in the same composite as the flood and runoff figures, so moving it would separate an input from its rollup. **If the backend shows that composite doesn't actually read those fields, my reason is wrong** and I'd rather know now.

2. **Is refusing `modules.ts` for the responsibility four correct?** `aec_api/routers/responsibility.py`'s own docstring says the rows *are* ordinary `responsibility` module records whose per-cell CRUD goes through the generic `/modules` routes — so storage, adjacency and the backend's own wording all point at `modules.ts`. I filed them separately on the grounds that storage is a HOW, not a question. That is a judgment call and I'd like it contested if you disagree.

3. **Mechanics, same as last time:** the mixin chain now has 40 wrappers after two more string-surgery insertions; and the extraction script walks back over contiguous `//` lines, so please check for a banner left orphaned in `client.ts` or a method that arrived in a new file without its comment. I removed the `ResponsibilityMatrix` type import after `tsc` showed it had no reader left — worth confirming nothing else did.

---

ibuilder:
Following up on my own question 1 before the review lands — I checked the backend rather than leaving a falsifiable claim hanging.

**The claim holds.** `resilience.climate_risk()` (`services/api/src/aec_api/resilience.py:234`) reads exactly the fields I said it does:

```python
wx = exposure or _weather_exposure(db, pid)
...
if wx.get("high_severity_open"):        score += 2
elif wx.get("open_risk_count"):         score += 1
if (wx.get("weather_delay_days") or 0) >= 5:  score += 1
```

They are scored alongside `flood.in_special_flood_hazard_area`, `flood.at_risk_count` and `storm.peak_runoff_cfs` into one rating. So the composite genuinely does consume the weather half, and separating `resilienceWeather` from it would split an input from its rollup.

**But there is a refinement that sharpens the tension rather than dissolving it.** `_weather_exposure` carries this comment in the source:

```python
wx = exposure or _weather_exposure(db, pid)   # site-risk + delay only — skips the schedule_activity scan
```

So the composite reads the **site-risk and delay** half of `resilienceWeather`'s payload and **not** the `weather_sensitive_activities` half — which is precisely the part that reads like `schedule.ts`'s (trade, start, finish, percent complete). My reason binds the half of that method that isn't schedule-shaped; the schedule-shaped half has no claim from the rollup at all.

That doesn't change the placement — it is one method and can't be split — and the method as a whole answers "what is the weather doing to us". But it is the honest version of the argument, and if `resilienceWeather` is ever separated, the activities list is the half with a claim elsewhere. Worth knowing before someone reads the header's argument as stronger than it is.

Please still contest the placement if you read it differently, and question 2 (`modules.ts` for the responsibility four) is the one I consider genuinely open.

---

ibuilder:
@coderabbitai Your framing was better than mine and it's now in the file — `9ad7eb3e`.

You made a distinction I hadn't: shared source metrics is **not** an endpoint-to-endpoint dependency. My header claimed the stronger thing, and the source only supports the weaker one. `climate_risk` reaches those fields through `_weather_exposure()`, which the backend comments as *"site-risk + delay only — skips the schedule_activity scan"* — so the rollup binds the half of `resilienceWeather` that is **not** schedule-shaped, and the schedule-shaped half has no claim from it at all. That is exactly backwards from what a reader would infer from "separating an input from its rollup".

And your reason is a better one than the dependency ever was, because it generalises:

> `schedule.ts` should contain programme-state queries, not every query that reads schedule activity records.

That is a boundary rule the next slice can apply; a dependency is a fact about two functions today. The header and changelog now carry that rule as the reason, with the narrowed dependency recorded beside it at its real strength.

No code moved — the placement is unchanged and the extraction is untouched. This corrects the recorded reasoning only. It cost a CI cycle, which I think is right: a header is where an argument gets inherited, and this whole sequence exists because prose beside an assertion drifts. Leaving a reason I knew to be overstated in a file I had just written would have been the same defect, authored deliberately.

Noted on point 3 as well — I hadn't registered that portal code reaches `ResponsibilityMatrix` through `client.ts`'s re-export of `./types`, so dropping the named import was safe for a reason beyond the one `tsc` gave me.

---

PR #408, opened 2026-09-04
SCALE-SEAM (91) — the R20 CRE deal desk, and a forecast tested instead of executed

ibuilder:
# What & why

Thirteen methods out of `apps/web/src/api/client.ts` (`867 → 746`) into a new `apps/web/src/api/creDeal.ts`, lifted as one contiguous run. A mixin, so every call site resolves unchanged. Roadmap item **SCALE-SEAM**, which stays open: 88 methods still sit above the STAYING banner with no map.

**What they answer: *should we transact on this income property, and on what terms?*** Verify the seller's numbers (`normalizeT12`, `rentRollScrub`, `netEffectiveRent`, `tieredComps`, `competitiveSupply`), decide (`holdSell`, `decisionGate`), set the terms (`clausePlaybook` + `reviewContractClauses`, `loanCovenants`, `dealAuthority`, plus the two save halves).

### Three signals, derived independently, and they agreed

1. **The source marks them.** Ten `CRE-` codes on these doc comments — and grepping all of `apps/web/src` showed those eleven occurrences were the **only** ones in the web tree, every one in `client.ts`.
2. **A 1:1 router match.** Every route these thirteen call (eleven distinct paths — the playbook and the authority table each have a read and a write half on one path) is served by `services/api/src/aec_api/routers/realestate.py` and by no other router; its ten `(R20)` docstrings name the same ten codes.
3. **One question runs through all thirteen.**

A marker alone would not have been enough — four shared-word traps (entitlements, view, carbon, lifecycle) are why `CRE-` counts as one vote of three and the router match carries the weight.

### A fourth corroboration, found after the slice was committed

The backend suite partitions the same thirteen: `test_cre_deal_desk`, `test_cre_governance`, `test_cre_tier3` and `test_net_effective` cover all eleven paths, and no `test_cre_*` file **asserts on** a route outside the set.

*Caveats, because this arrived late and is the kind of evidence it is tempting to round up.* `test_cre_tier3` reaches **three** routes outside the set as fixtures — `POST /proforma/scenarios`, `POST /proforma/scenarios/{sid}/review` and `GET /projects/{pid}/reports/ic_memo.pdf` — and `test_net_effective` sets up through `/rent-roll` and `/modules/lease`. The first draft of this claim named one of those three and truncated its path; review caught it. A test may use a neighbour as a fixture without that neighbour joining the cluster, but the count has to be right.

It was noticed only because two of those test names scrolled past in the suite output — test-file names are a grouping somebody else authored, and they were not on the list of places this slice thought to look.

### The boundary that decides `rent-roll`

`proforma.ts` already holds `GET /projects/{pid}/rent-roll`, and two of these sit directly beneath it on `/rent-roll/scrub` and `/rent-roll/net-effective` — a prefix split takes all three. The backend states the boundary: the plain rent roll carries no `CRE-` code and no `(R20)`, and its docstring calls it the operating rent roll "from the `lease` module (**the hold phase**)", while these two POST figures the counterparty supplied.

### (88)'s forecast was tested rather than executed, and it lost

(88) left `camReconciliation` with a note saying it "goes with `rentRollScrub`, `netEffectiveRent` and `normalizeT12` … when a rent-roll slice takes them". This is that slice; it took all three and left `camReconciliation` where it was — no `CRE-` code, no `(R20)`, and `/projects/{pid}/cam/reconciliation` is served by `operations.py`. **A placement forecast written by one slice is a hypothesis for the next slice to test, not an instruction to carry out.**

### The review also found the no-behaviour-change claim resting on a slack gate

This body originally said `apps/web/src/api/surface.test.ts` proves every call site still resolves. For these thirteen it did not. That file spot-checks a name list none of them was on, plus a floor of **751** on the total client surface — and the surface **measures 788**. Losing one of the thirteen leaves 787 and passes; losing the whole mixin leaves 775 and passes.

That is the decay `surface.test.ts` predicts about itself beside its previous 696 floor: *"a merge that adds an endpoint silently converts this ratchet back into a slack floor."* All thirteen are named there now, on the same reasoning the file already gives for the twenty auth methods, and mutation-checked both ways: renaming `holdSell` fails with *"holdSell() vanished — a call site is now broken"*; restoring it passes. **The fix is to name the group, not to raise the floor** — raising it restores the ratchet for one slice and leaves the next one in the same position.

Also corrected: the STAYING banner cited `rentRollScrub` as evidence of unfinished work (second time it has named a departed method), and two stale roadmap numbers — `client.ts is 953 lines`, three slices out of date, and a lane-table cell still saying (81)–(87).

- [x] Backend: affected `test_*.py` pass — full suite **660/660** (947s); `test_file_sizes.py`, `test_claude_md_gates.py`, `test_roadmap_status.py`, `test_ruff_scope.py`, `test_declared_imports.py` each exit 0
- [ ] Version bumped in both manifests — **n/a**, not a release PR; this is the thirtieth follow-on on `v0.3.1143`, whose tag lag is already at the `test_release_current` bound

Extraction ratchet lowered with the file: `867 → 746`.

## Summary by CodeRabbit

- **Features**
  - Continued support for commercial real estate deal workflows, including hold/sell analysis, contract reviews, loan covenant checks, deal authority, competitive supply, decision gates, comparable analysis, T12 normalization, rent-roll scrubbing, and net-effective rent calculations.
  - Existing API capabilities remain available with no behavioral changes.

- **Documentation**
  - Updated roadmap and changelog information to reflect current API coverage and supported deal workflows.

- **Reliability**
  - Added checks to ensure commercial real estate methods remain accessible and functional.


ibuilder:
@coderabbitai review

Worth a close look at two things in particular:

1. **The completeness of the cluster.** The claim is that thirteen methods form one question and that nothing else in `client.ts` belongs with them. It rests on four signals — ten `CRE-` doc codes that occur nowhere else in `apps/web/src`, a 1:1 match to `realestate.py` and no other router, its ten `(R20)` docstrings, and four `test_cre_*` backend test files that partition the same eleven paths. If any of those is weaker than stated, the header overstates and should be corrected rather than trimmed.

2. **The `rent-roll` boundary.** `apps/web/src/api/proforma.ts` already holds `GET /projects/{pid}/rent-roll`, and this PR puts `/rent-roll/scrub` and `/rent-roll/net-effective` in a different file. The argument is that the plain one is the hold-phase operating view (no `CRE-` code, no `(R20)`, its docstring says so) while these two take the counterparty's figures. That is a deliberate split of a route prefix across two mixins — the reverse of the usual instinct — so it is the decision most worth challenging.

The extraction body is byte-identical to what was removed; `apps/web/src/api/surface.test.ts` is what holds the no-behaviour-change claim.

---

ibuilder:
Both addressed in `66b85c04`. The documentation finding was right, and chasing it down turned up something larger that the review had the wrong way round.

**The caveat did undercount — fixed.** `test_cre_tier3.py` reaches three routes outside the eleven-path set, not one: `POST /proforma/scenarios`, `POST /proforma/scenarios/{sid}/review` and `GET /projects/{pid}/reports/ic_memo.pdf`. My caveat named only the last and truncated its path to `/reports/ic`. All three are fixtures rather than assertions, so the claim is now stated as *"no `test_cre_*` file **asserts on** a route outside the set"* — weaker than the first draft, and the one the evidence actually supports. Corrected in the mixin header, `CHANGELOG.md`, `docs/roadmap.md` and the size-pin comment.

**On `surface.test.ts` — the review's reassurance is off by 37, and my header was overstated in the same direction.**

> It does not list these methods by name, but losing any one reduces the count below 751.

I measured the runtime surface rather than reading the floor: it is **788**, against `expect(surface.size).toBeGreaterThanOrEqual(751)`. Losing one of the thirteen leaves 787 and passes. Losing the whole mixin leaves 775 and passes. The count guarded nothing here — which means my own header sentence, *"`api/surface.test.ts` is what proves that"*, was claiming more than the file delivered.

That is exactly the decay `surface.test.ts` predicts about itself, in the comment beside its previous 696 floor: *"a merge that adds an endpoint silently converts this ratchet back into a slack floor."* Thirty-seven endpoints have been added since the floor was last raised, and this is the first time it has misled a reader.

So all thirteen are now named in that spot-check, on the same reasoning the file already gives for naming the twenty auth methods. Mutation-checked both directions: renaming `holdSell` fails with *"holdSell() vanished — a call site is now broken"*, restoring it passes.

**The fix is to name the group, not to raise the floor** — raising it would restore the ratchet for one slice and leave the next one in the same position. This is the third time SCALE-SEAM has hit a gate that decays without ever going red; the other two were `client.ts > 1200` and the banner-string proxy that would have declared the item finished with 126 methods outstanding.

Thanks for the pull on the corroboration claim — the caveat being one route short is what sent me to check the other one.

---

PR #409, opened 2026-09-04
SCALE-SEAM (92) — annotation, and the type that says why the recipes were stuck

ibuilder:
# What & why

Four methods out of `apps/web/src/api/client.ts` (`746 → 733`) into a new `apps/web/src/api/annotate.ts`: `addAnnotation`, `addDimension`, `addRevisionCloud`, `addTag`. A mixin, so every call site resolves unchanged, and all four are named in `apps/web/src/api/surface.test.ts` — which (91) established is necessary rather than optional. Roadmap item **SCALE-SEAM**, still open: 84 methods above the STAYING banner.

**What they answer: put a note, a dimension, a cloud or a tag ON the model.** All four author real `IfcAnnotation` entities at world `[E,N]` through `editIfc`.

### The set is bounded by two sources this extraction did not author

1. **`services/api/src/aec_api/authoring_matrix.py`** — the curated recipe→category map behind the public authoring-coverage endpoint — has an `annotate` category holding **exactly these four** of its 99 recipes across 15 categories. A complete category: nothing left behind, nothing pulled in.
2. **`services/api/test_annotation.py`** exercises those four recipes and no others.

Third corroboration: `apps/web/src/viewer/tools/annotationSection.ts`, the UI these serve, makes exactly four `api.*` calls and they are these four.

### The `UX-2` marker does NOT bound the set, and the distinction is the point

(91)'s `CRE-` did — it occurred only in `client.ts`, so it drew its own boundary. `UX-2` occurs 12 times across three files and marks a **feature workstream** spanning UI and client, so it corroborates purpose and is silent on membership. Two markers, two strengths — and (91)'s success with `CRE-` is exactly what would make treating this one the same way feel justified.

### Not `markup.ts`, and not `authoring.ts`

`markup.ts` is **2D sheet markup** — a pin at a sheet's `(x, y)` carrying a note, promotable to an RFI. These author **model content** that travels with the IFC.

`authoring.ts` is the harder call: it **defines** `editIfc`, holds seven recipes, and claims *"the endpoints that WRITE to the model rather than read from it"* — which would take all 24 recipes still in `client.ts`. **A description broad enough to take everything is not a seam.** Its practice declined detailing at ⓼, property-override layers at ㊳, and groups at ⓻. `mep.ts` settled the principle already (*"use `editIfc` (`/edit`) and stay"*), and the matrix proves it: those 24 recipes split across **nine** categories.

### A technical finding that probably explains the whole backlog

**This is the first mixin to call `editIfc` without declaring it.** `editIfc` sits on the `Authoring` mixin rather than on `HttpCore`, so a mixin typed `Ctor<HttpCore>` cannot see it — the extraction does not compile, and nothing says why. `authoring.ts` calls it seven times but *defines* it, so it never needed a base type promising it, which is how eleven slices passed these recipes without meeting this.

`annotate.ts` declares the requirement in its type (`NeedsEditIfc`) and must be composed outside `withAuthoring`. Mutation-checked: composing it *inside* fails with `TS2345` naming `editIfc`, rather than at runtime with `undefined is not a function`. **Moving `editIfc` down into `HttpCore` would unblock the other 20 recipes** — a change to a file every mixin depends on, left for the next recipe slice.

*A related blocker is not the same blocker:* `authoring.ts` records *"a mixin cannot see a sibling's private member"* for its SSE methods. That one is harder — `liveStream` is `private`, unreachable at runtime too, and genuinely has to move. `editIfc` is public, so only the declared base **type** is missing it, which is why a type fixes this one and could not fix that one.

- [x] Backend: affected `test_*.py` pass — `test_file_sizes.py`, `test_claude_md_gates.py`, `test_roadmap_status.py` each exit 0 against the final tree. **The full-suite run was started before the last round of comment corrections, so it is a cross-check rather than the authority here**; CI's API test gate runs the final tree, and I'll report its result.
- [ ] Version bumped in both manifests — **n/a**, not a release PR; thirty-first follow-on on `v0.3.1143`, whose tag lag is at the `test_release_current` bound

Extraction ratchet lowered with the file: `746 → 733`, set to the measured line count rather than compressing the comment to fit a number I chose.

---

## Summary by CodeRabbit

* **New Features**
  * Added API support for creating annotations, dimensions, revision clouds, and tags.
  * Annotation tools support geometry, text, host elements, optional metadata, and publishing settings.

* **Refactor**
  * Reorganized annotation capabilities within the API client without changing their availability to end-users.

* **Documentation**
  * Updated the development roadmap with details about annotation API coverage and test scope.


ibuilder:
@coderabbitai review

Three things are worth adversarial attention here, in descending order of how much I'd like to be wrong:

1. **The `NeedsEditIfc` type and the composition-order requirement.** `withAnnotate` must be composed outside `withAuthoring` because `editIfc` lives on the `Authoring` mixin, not `HttpCore`. I've mutation-checked that composing it inside fails with `TS2345`, but please check whether the declared signature actually matches `authoring.ts`'s `editIfc` — a signature that is subtly *wider* than the real one would typecheck here and still be wrong at a call site.

2. **The completeness claim.** The set is asserted to be bounded by `authoring_matrix.py`'s `annotate` category (exactly four of 99 recipes across 15 categories) and by `test_annotation.py` (those four recipes, no others). If either is weaker than stated — a fifth recipe categorised `annotate` somewhere I didn't look, or a test touching a fifth — the header overstates and should be corrected rather than trimmed.

3. **The placement argument against `authoring.ts`.** That file *defines* `editIfc` and describes itself as "the endpoints that WRITE to the model", which would take all 24 remaining recipes. I argue its demonstrated practice is narrower than its sentence (it declined detailing at ⓼, property-override layers at ㊳, groups at ⓻) and that the matrix splitting those 24 across nine categories proves `editIfc` isn't the seam. That's a judgement call and the one most worth pushing back on.

The extraction body is byte-identical to what was removed, and all four methods are named in `surface.test.ts` — which the previous slice established is necessary because that file's count is a floor with 37 of slack.

---

ibuilder:
Both addressed in `5a79f66b`. Thanks especially for checking the signature — that was the gap I couldn't close myself.

**On `NeedsEditIfc`:** confirming your finding independently, all five parameters match `Authoring.editIfc`, `publish?: boolean` correctly represents `publish = true`, and the return type is identical. That's the thing my own mutation test *couldn't* prove: composing the mixin in the wrong order fails loudly, but a signature subtly **wider** than the real one would have typechecked cleanly here and broken at a call site instead. Verified, not just accepted.

**On `test_annotation.py`:** you're right, and I verified it directly — `edit.add_wall` and `edit.add_column` are each called once, as hosts for `add_tag`. Corrected to "exercises all four annotation recipes and no other recipe **as its subject**", in the mixin header, `CHANGELOG.md`, `docs/roadmap.md` and the size-pin comment.

**Worth naming: this is the second consecutive slice with that exact error.** On #408 I claimed no `test_cre_*` file reached a route outside its set; `test_cre_tier3` reached three as fixtures. Both times the claim counted what a test *targets* and ignored what it *sets up*, and both times a reviewer found it rather than I did.

So it's recorded as a rule rather than a typo: **a "nothing else" claim about a test has to say whether it means the assertions or every call the file makes.** Those are different sets, the smaller one is the one you notice while reading for what a test is *about*, and a corroboration argument resting on the smaller set while claiming the larger one is overstated in precisely the direction that makes a cluster look tidier than it is.

Re-verified on this head: `tsc` 0, eslint 0, vitest `src/api` 0, `ruff check ../..` clean, and `test_file_sizes` / `test_claude_md_gates` / `test_roadmap_status` all exit 0.

---

ibuilder:
All four findings resolved. Three of them were already fixed before the review landed — it ran against `2bcfe5e8`, and `5a79f66b` had corrected the `test_annotation.py` qualification in the mixin header, `docs/roadmap.md` and the size-pin comment. **The fourth was live and I had missed it**, now fixed in `65491dda`.

**`CHANGELOG.md:980`** still opened the entry with `746 → 732` while line 1048 and every other record said 733. Good catch — and the way I introduced it is worth naming.

When my banner correction pushed `client.ts` from 732 to 733, I updated the *pin* and the *"Pin 746 → 733"* line, and stopped there. I didn't re-derive every place the entry states that number. **That's the same residue problem the fix was for, applied to the fix itself** — a correction has the same reach as the claim it corrects, and I checked one of its two sites.

So I verified this one by grepping for `746 -> 732` and `746 → 732` across `.md`/`.py`/`.ts` rather than by re-reading the paragraph I'd just edited. Zero remaining.

**On the Docstring Coverage pre-merge warning (0%, threshold 80%):** not acting on it. It analysed one function — `withAnnotate` — which sits directly beneath a 60-line header block explaining the extraction, the boundary and the `NeedsEditIfc` contract. The four extracted methods carry their original `/** UX-2: … */` docs, moved byte-identically. Adding a JSDoc block above the factory to satisfy a percentage would duplicate what's already immediately above it. The repo's own documentation gates (`test_claude_md_gates.py`, `apps/web/src/api/docComments.test.ts`) pass. Happy to reconsider if you think the factory specifically warrants its own tag block.

Re-verified on `65491dda`: `test_claude_md_gates` and `test_file_sizes` both exit 0.

---

PR #410, opened 2026-09-04
SCALE-SEAM (93) — finishing ⑲, and a forecast of mine that did not survive testing

ibuilder:
# What & why

Two methods out of `apps/web/src/api/client.ts` (`733 → 731`) into the **existing**
`apps/web/src/api/mep.ts`: `connectMep` and `addMepFitting`. No new mixin. Roadmap item
**SCALE-SEAM**, which stays open: 82 methods still sit above the STAYING banner with no map.

### ⑲ already claimed them, and could not take them

`mep.ts` has said since SCALE-SEAM ⑲ that *"`connectMep` / `addMepFitting` call `editIfc` (`/edit`)
and stay."* That reads as a placement decision. It was a **compiler error**: `editIfc` is declared on
the `Authoring` mixin rather than on `HttpCore`, so a mixin typed `Ctor<HttpCore>` cannot see it and
the extraction does not compile. (92) diagnosed that; this applies it.

**A method left behind for a reason nobody wrote down stays behind indefinitely.** ⑲ recorded what
happened and not why, so eleven slices read its note as settled and moved past.

### One contract, not two copies

`NeedsEditIfc` moves from `annotate.ts` into `apps/web/src/api/types.ts`, shared by both mixins. Two
hand-copied signatures are exactly the drift the #409 review had to check by hand — a copy subtly
*wider* than the real `editIfc` typechecks at the mixin and still breaks at a call site. One
definition cannot drift from itself. **Mutation-checked:** composing `withMep` inside
`withAuthoring` fails with `TS2345` naming `NeedsEditIfc`; restoring passes.

### A forecast I shipped one slice ago, tested and withdrawn

(92) recorded in four places that *"moving `editIfc` down into `HttpCore` would unblock the other 20
recipes"*, and left it as the recommended next step. **It does not hold.** `httpCore.ts` says it owns
"the base URL, the bearer token, and the low-level fetch helpers", keeping *"transport concerns
separate from the endpoint surface"* — and all eight of its methods are transport. `editIfc` is a
**domain** endpoint (`POST /projects/{pid}/edit`). Moving it there contradicts that file's stated
purpose and the layering SCALE-SEAM exists to build. All four records are corrected.

I wrote that forecast as a plan **one slice after (91) recorded that a placement forecast is a
hypothesis for the next slice to test, not an instruction to carry out.** Phrasing it as a plan is
precisely what makes it read as settled — the failure (91) named, committed by the slice that named it.

### The evidence is weaker than (92)'s, and says so

**Not** a complete `authoring_matrix.py` category: `create-mep` has 11 members and `edit-mep` 3, of
which these are one each. The boundary is *"MEP recipe exposed as a **typed client method**"*, which
is exactly these two — the other nine (`add_duct`, `add_pipe`, `add_riser`, `add_sprinkler`,
terminals, devices) are driven from viewer code through the generic recipe path in
`apps/web/src/viewer/draft/draftCatalog.ts` and `apps/web/src/viewer/tools/mepSection.ts`, so they
were never client methods to move. `services/api/test_mep_systems.py` covers both **among** other MEP
recipes — corroborating the family, not bounding the set — and `test_guards.py` is cross-cutting.
*Spelled out because the "nothing else" overstatement was made in each of the two preceding slices.*

`addMepFitting` is callerless and stays on `apps/web/src/api/clientCallers.test.ts`'s allowlist:
moving a method does not change its called-ness. Both are still named in `surface.test.ts`, because
that list is about the **surface** rather than the call sites — a method with no screen behind it can
still vanish in an extraction, and the count would not notice.

- [x] Backend: affected `test_*.py` pass — `test_file_sizes.py`, `test_claude_md_gates.py`, `test_roadmap_status.py`, `test_ruff_scope.py`, `test_declared_imports.py` each exit 0; full suite running as a cross-check, CI's gate is the authority for this head
- [ ] Version bumped in both manifests — **n/a**, not a release PR; thirty-second follow-on on `v0.3.1143`, whose tag lag is already at the `test_release_current` bound

Extraction ratchet lowered with the file: `733 → 731`.

---

## Summary by CodeRabbit

* **New Features**
  * Added MEP authoring actions for connecting MEP elements and adding MEP fittings.
  * Added optional fitting details, including predefined types, sizes, and systems.
* **Bug Fixes**
  * Improved compatibility between MEP authoring operations and editing capabilities.
* **Documentation**
  * Updated the roadmap and changelog with the latest API changes and MEP recipe coverage.


ibuilder:
@coderabbitai review

Three things worth adversarial attention, in descending order of consequence:

1. **`NeedsEditIfc` moving to `types.ts`, now with two consumers.** It's the same shape you verified on #409, but it's shared now — please check `annotate.ts` and `mep.ts` both actually resolve it from `./types` with no leftover local copy, and that the signature still matches `authoring.ts`'s `editIfc` exactly. A shared contract that drifts from its implementation is worse than two copies, because it looks authoritative.

2. **The composition-order requirement, now on a second mixin.** `withMep` must compose outside `withAuthoring`. I mutation-checked it (composing inside → `TS2345`), but `withMep` sits much deeper in the chain than `withAnnotate` did, so it's worth confirming the chain genuinely satisfies it rather than passing by accident.

3. **The set boundary, which is deliberately weaker than the last slice's.** These two are *not* a complete `authoring_matrix.py` category. The claim is "MEP recipe exposed as a typed client method" — exactly two, with the other nine driven from viewer code via the generic recipe path. If you find a tenth MEP recipe that *does* have a typed client method, the boundary is wrong and the header overstates.

Please also apply the check that caught me on both preceding slices: I've claimed `test_mep_systems.py` covers these two **among** other MEP recipes and therefore does *not* bound the set. That's the deliberately weaker phrasing — but check I've drawn the line in the right place this time rather than just hedging it.

The two moved bodies are byte-identical to what was removed, and both are named in `surface.test.ts`.

---

ibuilder:
Fixed in `47389a4e`. You're right, and the arithmetic is the least of it.

**The count.** 11 `create-mep` + 3 `edit-mep` = 14, less the two moved = **twelve**, verified by parsing the matrix rather than re-reading my own sentence. Nine was the number I had *enumerated* — the viewer-driven ones — and I wrote it as "the other nine", which asserts a complement I never computed.

**The three I'd silently dropped deserve their own line.** `add_sprinkler`, `auto_connect_mep` and `set_system_predefined` are referenced **nowhere in `apps/web/src`** — backend MEP recipes with no web exposure at all. Recorded, not fixed; wiring a capability isn't an extraction slice's business. But I want to be clear that I wouldn't have found them by being more careful. Only by computing the remainder.

**Because this is the third consecutive slice with one defect wearing a third disguise:**

| | claim | reality |
|---|---|---|
| #408 | no `test_cre_*` file reaches outside the set | three reached out as fixtures |
| #409 | `test_annotation.py` exercises four recipes "and no others" | two more as tag-host fixtures |
| #410 | "the other nine" | a subset enumerated, phrased as the complement |

Each time: **count the members you looked at, then phrase the result as the whole population.** You found all three. I found none of them by re-reading, which is the part worth taking seriously — re-reading confirms the sentence I wrote, not the set it describes.

So the correction recorded in the code isn't "be more careful". It's **derive the complement** — `set(all) - set(moved)` — and only then describe what's in it. All three of those would have died on that one line.

That's the same lesson this repo's gates already encode, arriving in prose: a claim computed over the examples you gathered reads exactly like a claim computed over the population, and only the derivation distinguishes them. I've been writing that sentence about other people's checks all session while doing it myself in three consecutive headers.

Corrected in `mep.ts`, `CHANGELOG.md`, `docs/roadmap.md` and the size-pin comment. Re-verified: `tsc` 0, eslint 0, vitest `src/api` 0, `test_file_sizes` / `test_claude_md_gates` / `test_roadmap_status` all 0.

---

ibuilder:
All three inline comments are the same finding, and it's already fixed in `47389a4e` — the review ran against `09299541`, the pre-fix head.

**I took the stronger of the two options you offered.** Your suggestion was to drop the count ("Other MEP authoring recipes use the generic viewer recipe path or remain outside the typed client surface"), with an alternative: *"or document the complete set of remaining recipes."* I did the latter, and verified each member individually rather than asserting a total:

| recipe | where |
|---|---|
| `add_duct`, `add_pipe`, `add_cable_tray`, `add_wire`, `add_mep_terminal` | `viewer/draft/draftCatalog.ts` |
| `add_riser`, `add_fire_equipment`, `add_fa_device`, `add_comms_device` | `viewer/tools/mepSection.ts` |
| `add_sprinkler`, `auto_connect_mep`, `set_system_predefined` | **not referenced anywhere in `apps/web/src`** |

14 in the matrix − 2 moved = **12**, split 9 / 3. The two you named are in that third row.

I kept the count rather than removing it because **the count is the part that was doing work.** "Other recipes use the generic path or are outside the typed surface" is unfalsifiable — it can't be wrong, so it can't be checked, and the next slice inherits no information. The enumerated version is checkable and it immediately surfaced something I'd have otherwise shipped past: three backend MEP recipes with no web exposure at all. Recorded, not fixed — wiring a capability isn't an extraction slice's business.

That's also why I'd push back gently on the general shape of the suggestion: **the fix for an overstated count is a derived count, not a vaguer sentence.** Weakening the prose would have hidden the same gap it was hiding before, just less visibly.

`47389a4e` also records why this kept happening — three consecutive slices, one defect: count the members you looked at, phrase the result as the whole population. The rule now written into the code is to derive `set(all) − set(moved)` first and describe it second.

---

PR #411, opened 2026-09-04
SCALE-SEAM (94) — the two methods ㊻ could not see, and a complete category that is not one question

ibuilder:
# What & why

Two methods out of `apps/web/src/api/client.ts` (`731 → 727`) into the **existing**
`apps/web/src/api/model.ts`: `verifyAsbuilt` and `recordAsbuiltDimension`. No new mixin. Roadmap item
**SCALE-SEAM**, which stays open: 80 methods still sit above the STAYING banner with no map.

### ㊻ stated this question and left two of its members behind

`model.ts` has said since SCALE-SEAM ㊻ that it holds *"field-install verification — **is this element
installed as designed?**"* — coverage, status write, deviation log. `verifyAsbuilt` stamps
`Massing_AsBuilt` (the LOD-500 reliability layer); `recordAsbuiltDimension` records a field-verified
dimension with its variance against design. Both answer exactly that question, and **neither is named
in ㊻'s header.**

**Why it missed them: ㊻ split on the `/verification/*` route prefix, and these two go through
`editIfc`, so they carry no route to be found by.** A route-prefix split cannot see a recipe-based
sibling of the same question. They were also unmovable at the time, for the reason (92) diagnosed —
`editIfc` lives on the `Authoring` mixin, so a mixin typed `Ctor<HttpCore>` does not compile. That
makes **three** slices now known to have left work behind for that one cause, after ⑲'s MEP pair.

### The matrix argued for a third method, and it was declined

`services/api/src/aec_api/authoring_matrix.py`'s `lifecycle` category is exactly these two plus
`set_phase` — a **complete category**, the signal that decided (92)'s annotate slice. `setPhase`
stays anyway, on two checkable grounds:

- it answers a different question — tag a construction phase (new / existing / demolish / temporary),
  which is element state in the build sequence, not install-versus-design; and
- **its read half `phasing()` is still in `client.ts`.** Taking a writer and leaving its reader is the
  split (87) had to undo.

*A complete category is one vote, not a verdict — and "lifecycle" is a word (88) already caught
misleading once, in a different way.*

### Two checks that failed honestly

**The first composition-order check was worthless, and reading its output is the only reason I know
that.** It rebalanced parentheses wrongly and went red with `TS1005` syntax errors — proving nothing
about composition order while looking exactly like a passing mutation test. *A check that goes red
for a reason other than the one claimed is the same defect as one that goes green for it.* Redone as
an isolated probe applying `withModel` to bare `HttpCore`, which fails `TS2345` naming `NeedsEditIfc`.

**The ratchet caught the banner.** The first draft added 9 lines of prose to `client.ts` while the
extraction removed 8, netting **+1** — an extraction slice that grows the file has failed its purpose.
Condensed to three lines, with the reasoning living once in `model.ts`'s header instead of twice.
*Not (91)'s "don't shave prose to fit a pin": there I compressed an accurate comment to clear a number
I'd set myself; here the file grew, and the duplicated paragraph was the right thing to cut.*

- [x] Backend: affected `test_*.py` pass — `test_file_sizes.py`, `test_claude_md_gates.py`, `test_roadmap_status.py` each exit 0; full suite running as a cross-check, CI's gate is the authority for this head
- [ ] Version bumped in both manifests — **n/a**, not a release PR; thirty-third follow-on on `v0.3.1143`, whose tag lag is already at the `test_release_current` bound

Extraction ratchet lowered with the file: `731 → 727`.

---

## Summary by CodeRabbit

* **New Features**
  * Added API support for recording as-built dimensions and verifying as-built information.
* **Documentation**
  * Updated the changelog and roadmap with the latest as-built API details and clarified scope.
* **Refactor**
  * Reorganized as-built operations within the API without changing their available capabilities.
* **Tests**
  * Added compile-time checks to enforce valid API mixin composition and prevent unsupported configurations.


ibuilder:
@coderabbitai review

Three things worth adversarial attention, in descending order of consequence:

1. **The declined third method — this is the judgement call.** `authoring_matrix.py`'s `lifecycle` category is exactly `verify_asbuilt`, `record_asbuilt_dimension` and `set_phase`. That's a *complete category*, and a complete category is the signal that decided the previous annotate slice. I've taken two and left `set_phase`, arguing it answers a different question and that its read half `phasing()` is still in `client.ts`. If you think the category should have been taken whole, say so — I'd rather be argued out of this now than have it read as settled by the next slice.

2. **`withModel` now requires `NeedsEditIfc`, which changes an existing mixin's base type.** Please confirm it genuinely composes outside `withAuthoring` in the chain, and that nothing else in `model.ts` depended on the `HttpCore` import I removed. My first check of this was invalid — it produced `TS1005` syntax errors rather than a type error, so it proved nothing while looking like a pass. The replacement is an isolated probe; worth confirming it tests what I claim.

3. **The completeness claim, stated deliberately narrowly this time.** I claim only that these two are members of ㊻'s question that ㊻ did not name, and that ㊻ missed them because it split on a route prefix they don't have. I am *not* claiming they're the only such methods. If there are other `editIfc` recipes answering "is this installed as designed", the boundary is wrong.

On (3) especially: the last three slices each overstated a set by counting the members I'd looked at and phrasing it as the population. This one is scoped to avoid that, but please check I've drawn the line where the evidence actually is rather than merely hedging.

Both moved bodies are byte-identical to what was removed, and both are named in `surface.test.ts`.

---

ibuilder:
All three addressed in `491c71df`. You confirmed two decisions and found the defect I'd asked about — which is the one worth dwelling on.

**The overstatement is real and it's the specific thing I asked you to check.** I scoped the claim carefully in the PR description — *"two known missed members, not necessarily all members"* — and then wrote **"finishes that question"** in `model.ts` and **"completing its verification question"** in `surface.test.ts`.

**Hedging in the review request while the artifact overstates is not scoping.** It puts the caveat where a reviewer sees it and the overstatement where every later reader will — worse than a plain overstatement, because it looks like discipline. Fourth instance of one defect across four slices, and the first where the correct wording *already existed* and simply didn't reach the code. Both now read "two members of ㊻'s question that it did not name".

**On the probe you couldn't verify — you were right that you couldn't, and that's a finding in itself.** I reported the constraint as "mutation-checked", but the check was a scratch file I'd deleted. **A verification that leaves no artifact is a claim, not a check.** So it's now `apps/web/src/api/compositionOrder.test.ts`: `@ts-expect-error` assertions that `withAnnotate`, `withMep` and `withModel` each reject a bare `HttpCore` base — the `NeedsEditIfc` requirement stated positively, enforced by `npm run typecheck`, naming the mixin when it breaks. Until now it was enforced only as a side effect of `tsc` succeeding on one very long expression in `client.ts`, which wouldn't have said *which* constraint broke.

Mutation-checked in turn: relaxing `withMep` to `Ctor<any>` fails with `TS2578: Unused '@ts-expect-error' directive` naming the line. The test's header also records the earlier scratch attempt, which went red with `TS1005` syntax errors — proving nothing about composition order while looking exactly like a passing mutation test.

**One correction to your read.** You described `phasing()`/`setPhase()` as a direct read/write pair in `modelStatePanels.ts`. `phasing()` is read there and the panel colours by `Massing_Phasing.Status` — but `setPhase` is that panel's **unwired** writer, sitting on `clientCallers.test.ts`'s UNCALLED allowlist. Reader and *unwired* writer on one panel. Same conclusion, and your `Massing_Phasing.Status` vs `Massing_AsBuilt`/`Massing_AsBuiltDim` distinction is a better argument for it than the one I'd written.

Re-verified: `tsc` 0, eslint 0, vitest `src/api` 27 files / 119 tests, `test_file_sizes` and `test_claude_md_gates` 0.

---

ibuilder:
The third site is fixed in `0b95af10`, and finding it is the useful part.

Your inline comments name three files. My previous fix (`491c71df`) corrected `model.ts` and `surface.test.ts` and **left `docs/roadmap.md:3238` saying "(94) finished SCALE-SEAM ㊻"** — the same overstatement, still live.

**That's the fifth instance of one defect, and the second consecutive PR where I corrected a *subset* of a phrase's sites.** On #410 I fixed the size pin but not the CHANGELOG sentence stating the same number. Here, two of three files. The rule I wrote for #410 — *derive the complement, don't enumerate the sites you remember* — applies to the fix as much as to the original claim: grep every occurrence of the wording, then edit. I didn't, twice running. It's now recorded that way rather than as another one-off correction.

**One thing I deliberately did not level.** `docs/roadmap.md` still says "(93) finished SCALE-SEAM ⑲", and I think that one is right:

- **⑲ named the two it left** — `mep.ts`'s header says `connectMep` and `addMepFitting` by name — so taking both completes an explicit list. "Finished" is defensible.
- **㊻ named nothing.** There's no list to complete, so no basis for claiming its question is whole — only that these two answer it and were missed.

Same word, different evidential footing. Blanket-editing both to the safer phrasing would have hidden that distinction, which is exactly the failure mode I flagged on #410 when declining to replace a wrong count with vaguer prose. The roadmap now says which is which.

Re-verified: `test_claude_md_gates`, `test_roadmap_status` and `test_file_sizes` all exit 0.

---

PR #412, opened 2026-09-04
SCALE-SEAM (95) — element state: two read/write pairs, and why the matrix loses

ibuilder:
# What & why

`apps/web/src/api/client.ts` goes **727 → 711** (76 methods above the STAYING banner, 4 below).
Four methods move to `apps/web/src/api/model.ts` — the `lodSummary`/`setLod` and `phasing`/`setPhase`
read/write pairs — because they answer one question: *what state are the model's elements in, and
set it?* Continues SCALE-SEAM in `docs/roadmap.md`.

**The grouping is derived, not asserted.** Five independent witnesses agree:

- identical return shape `{ total, <x>ed, prop, counts: Record<…> }`;
- both writers are `(pid, guids, <enum>, publish) → editIfc`;
- both readers are consumed by `apps/web/src/viewer/tools/modelStatePanels.ts` (lines 251, 316);
- both writers sit unwired and *adjacent* on `apps/web/src/api/clientCallers.test.ts`'s UNCALLED
  allowlist;
- `model.ts` already owned `/model/lod/census`, `/lod/handover-readiness` and `/lod/assessment`,
  while the base distribution `/projects/{pid}/lod` stayed behind — `lodSummary` was a sibling
  separated from its own family.

**`authoring_matrix.py` disagrees, and it is recorded as the losing vote rather than elided.** It
files `set_lod` under `data` and `set_phase` under `lifecycle`, because it categorises by the IFC
*output* each recipe writes — an LOD stage tag against `Massing_Phasing.Status`. Different property
sets, same question. That is (89)'s **"storage is a HOW"** trap: "they write different psets" has the
same shape as "they are all module records", and neither is a question. This is the first slice where
the matrix has been the losing vote after being right three running — worth naming, because three
straight wins is exactly how a corroborating source turns into an unexamined authority.

**It meets (94)'s objection rather than overriding it.** (94) declined `setPhase` because taking the
writer alone would have stranded `phasing()` in `client.ts` — the reader/writer split (87) had to
undo. Both halves move together here, so nothing is separated.

**Scope claim, stated at the strength the evidence supports:** these four answer that question and
belong together. This does *not* claim the question is now complete — no derivation of the complement
was done here, and the header says so.

The four names are added to `apps/web/src/api/surface.test.ts` because its floor is a *slack* ratchet
(788 actual vs a 751 floor), so the count alone would not notice a loss; and because the UNCALLED
allowlist is about **call sites** while that list is about the **surface** — an unwired writer can
still vanish in an extraction unnoticed.

- [x] Backend: affected `test_*.py` pass — `test_file_sizes.py`, `test_claude_md_gates.py`,
      `test_roadmap_status.py` all exit 0; full suite running as a cross-check
- [x] Web: typecheck, lint and build clean (`npx tsc --noEmit` = 0, `npm run lint` = 0,
      `npm run build` = 0); `npx vitest run src/api` = 27 files / 119 tests
- [x] `CHANGELOG.md` entry added (newest at top); `docs/roadmap.md` updated
- [ ] Version bump — n/a, not a release PR

---

## Summary by CodeRabbit

* **Refactor**
  * Moved element state tools for LOD stages and construction phases to the model API.
  * Existing summaries and editing actions remain available through the updated model interface.

* **Documentation**
  * Updated the roadmap and change records to reflect the API organization changes.

* **Tests**
  * Updated API surface checks to cover the relocated element state methods.


ibuilder:
@coderabbitai review

---

PR #472, opened 2026-09-08
CSV-SWEEP: every spreadsheet writer shares one cell guard, and numbers survive it

ibuilder:
# What & why

PR #471 fixed spreadsheet formula injection in the module CSV export and lifted the guard beside `roundtrip_export`. That fixed **two** writers; there are **nine**. This sweep derives the population instead of listing it, guards the six that were exposed, and fixes a regression #471 itself shipped. Roadmap: R22-ENTITLEMENT, under the ⑬ REF-LABEL entry.

**Six unguarded writers, each measured live rather than reasoned about:**

| writer | what reached the sheet |
|---|---|
| `accounting.to_gl_csv` | a vendor named `=cmd\|'/c calc'!A1` in the Vendor column of a general-ledger CSV an accountant opens |
| `aec_data.xlsx.write_sheets` | every XLSX export, the COBie workbook handed to an owner at turnover included — openpyxl stores such a string with `data_type='f'`, a **live formula in the file** |
| `model_query.to_csv` | element names out of an uploaded IFC |
| `layout.to_penzd_csv` | survey point descriptions |
| `drawing_schedules.schedule_csv` | door and window marks |
| `rebar_rules.bbs_csv` | bar marks, in the schedule a fabricator ingests |

The last two were found by the **derived gate**, not by the hand-written list — which is the argument for deriving a population rather than enumerating one. `+SUM(A1:A9)` is stored by openpyxl as `data_type='s'`, so XLSX exposure is narrower than CSV's; that was measured, not assumed.

**And the sweep found a regression #471 shipped.** That PR guarded every cell on the reasoning that *"a guard applied per-column is one column away from being forgotten"*. The reasoning is right and the guard was wrong: **a negative number begins with a formula lead**, so `cost_impact = -500.0` exported as `'-500.0` — a text cell where a number belongs, in a column somebody sums. Survey coordinates are the same shape and would have been corrupted outright. The guard is now number-aware: `-1+1` still escapes, `-500.00` does not, because the danger is an *expression* opening with a lead, not a signed value.

**#471's own test caught that, by pinning the wrong contract** — it asserted `csv_cell("-1") == "'-1"`, i.e. it demanded the defect. The lead-escaping property is unchanged; only the probe moved off values that are numbers, and two assertions replaced it (a plain number is not escaped; an expression opening with a lead still is).

**The guard now lives in `services/data/src/aec_data/cells.py`** — the *lower* package, so both packages share one definition rather than two spellings of a security rule. `aec_api.modules_query` re-exports it under the name the API already used.

**The gate is `services/api/test_csv_sweep.py`, and its predicate took four wrong drafts, each in the dangerous direction.** Bare `.append` matched ~200 `list.append` functions; adding `save` to disambiguate reported two reportlab PDF canvases; matching the openpyxl worksheet source alone reported two openpyxl *readers*, because an import path is not an export. The sink is now a `.writerow`/`.writerows` call, or a `.append` in a function that **both** holds a worksheet **and** saves it. Six shapes are probes the gate runs against itself **in both directions** — two authoring ones must be reported, four near-misses must stay silent — because a predicate that decides what to *look at* hides everything it excludes from its own output.

The QuickBooks IIF path is **deliberately left unguarded**, with the reason recorded beside it: IIF is consumed by QuickBooks, not a spreadsheet, so nothing evaluates a leading `=` and an apostrophe would corrupt the vendor name *on import*. Its real exposure is delimiter injection — a different class, and a different fix.

- [x] Backend: `cd services/api && python -m ruff check ../..` clean (whole tree)
- [x] Backend: affected `test_*.py` pass — **679/679 suites, exit 0**, on the final tree with nothing else running. `test_csv_sweep` registered in `run_tests.py` TESTS. Six mutations (three on the guard, three on the deriver) each caught by an assertion.
- [x] Web: not applicable — no files under `apps/web/` are touched
- [x] No import cycles introduced (`test_import_cycles.py` passes; `aec_api` → `aec_data` is the existing direction)
- [x] `CHANGELOG.md` entry added; roadmap ⑬ entry extended

---

## Summary by CodeRabbit

- **Security**
  - Spreadsheet CSV and XLSX exports now neutralize formula-like user text across supported formats.
  - QuickBooks IIF exports prevent tabs and line breaks in text from creating forged records.
  - Numeric values, including negatives, remain usable as numbers; expressions such as `-1+1` remain safely escaped.

- **Bug Fixes**
  - Fixed inconsistent protection across accounting, model, layout, drawing schedule, rebar, and spreadsheet exports.

- **Tests**
  - Added comprehensive coverage for export safety and numeric data handling.

- **Documentation**
  - Updated the changelog and roadmap with these improvements.


ibuilder:
Fixed in `a8af6838`. The finding was right, and it was right about something this PR had explicitly looked at and walked past.

CSV-SWEEP left `to_iif_bills` unguarded on purpose, with the reasoning in its docstring: QuickBooks doesn't evaluate a leading `=`, and prefixing an apostrophe would corrupt the vendor name *on import*. That half stands. **Stopping there was the mistake** — the same docstring named the delimiter exposure and then left it in the tree as "a different class."

**It's record forgery, not field corruption.** Measured on the pre-fix code, one hostile vendor name turned a single bill into **five records** (7 → 11 physical lines), writing columns the operator never entered into an accounting import nobody reads line by line.

`_iif` replaces each of `\t`, `\r\n`, `\r`, `\n` with a space:
- **replaced, not stripped** — `Acme<TAB>BILL` stays legible instead of becoming `AcmeBILL`;
- **replaced, not rejected** — a pasted multi-line address can't fail an export or silently drop a real bill.

Pinned by **count** rather than by content, in `services/api/test_csv_sweep.py`: a fixed number of input rows must yield a fixed number of physical lines *and* a fixed tab count per line, whatever the text contains. That's the property that can't be satisfied by accident. Three mutations are caught, including reverting `_iif` to the identity function.

Two things I got wrong writing the test, corrected before pushing: one assertion's second clause was a tautology (`"\t" not in FORGE.replace("\t", "")`), and the fixture used a negative `amount` where `journal()` supplies a positive invoice amount and `to_iif_bills` applies the AP sign itself.

679/679 backend suites, ruff clean over the whole tree.

---

PR #541, opened 2026-09-12
The blind spot was 70 routes wide, and two of them were written down

ibuilder:
# What & why

`LEAF_COLLISION_DARK` in `services/api/test_route_reachability.py` is a **triaged** list — routes
somebody read, confirmed dark, and recorded. The note above it described it as the output of
*"screening every route whose leaf this rule reads as called but which never appears path-shaped"*,
which is a different thing entirely. **A hand-curated list that describes itself as derived reads
exactly like a complete one.**

The actual population is every route the rule cannot see *at all*: leaf long enough to be checked,
leaf vouched by an ordinary English or HTML word (so the rule stays silent), and no caller that
builds the path. Derived rather than listed:

| | routes |
|---|---|
| total routes | 947 |
| flagged by the rule, frozen in `KNOWN_UNCALLED` with reasons | 56 |
| short leaf (its own existing ratchet) | 115 |
| leaf vouched, caller found | 696 |
| **leaf vouched, no caller — the blind spot** | **70** |
| …of which `LEAF_COLLISION_DARK` named | **2** |

## It is a record, not an allowlist

Frozen as `LEAF_COLLISION_BLIND`, on exactly the terms `KNOWN_UNCALLED` already states for the half
the rule *can* see. **70 is not a target and most of the set is probably correct:**

- `/scim/v2/Users` is for an identity provider, `/bcf/2.1/projects` for external BIM tools — neither
  should *ever* have a web caller.
- Some are **false positives of the probe itself**: `openDrawing("plan.dxf?…")` in
  `apps/web/src/viewer/tools/drawingsSection.ts` builds `/projects/{pid}/drawings/plan.dxf` through a
  helper that takes a relative suffix, so the literal never appears anywhere.

The value is not the number — it is that a number now exists. Before this the blind spot had no
measured size and nothing could report a new member. Frozen, it can only shrink, and route 71 arrives
with a name instead of silently.

**Triage is deliberately not attempted here.** No syntactic probe can classify these: the leaf being
vouched *is* the problem, and a leaf-based helper search returns 685 hits for `projects` and 44 for
`status` — the same collision one layer down. Each candidate needs a read, and that is its own work.

## Correcting the 26 I shipped an hour ago

PR #540's note put this at 26. That was measured with a looser probe asking whether the route's
**parent** fragment appeared (`/lod/`), which any sibling route satisfies — so a called
`/lod/assessment` vouched for an uncalled `/lod/matrix`, *the same collision one segment up*. 70 is
the count when the question is "does a caller build **this** path", which is the question that was
meant. Corrected in place in `CHANGELOG.md` rather than edited away, because it drifted in the
flattering direction and that is the hardest kind of drift to notice.

**Four measurements across two gates have now gone the same way** — 377 and 35 in `test_scratch_ignored.py`,
then 26 and 70 here — and the pattern is identical every time: *the checker did not fail, it
answered.* So the derivation carries two live controls it must separate.

## Mutation-tested

| mutation | result |
|---|---|
| drop `/scim/v2/Users` from the frozen set (simulates a 71st arriving) | red, naming the route |
| blind the derivation so it returns nothing | red on the **controls**, not a silently clean tree |

The controls are both live routes, so neither can pass by naming something that stopped existing:
`/bsdd/search`, wired an hour ago, must fall *out* of the set; `/scim/v2/Users`, callerless by design,
must stay *in*.

- [x] Backend: `cd services/api && python -m ruff check ../..` clean (whole tree) — "All checks passed!"
- [x] Backend: `test_route_reachability.py` passes standalone and was mutated in both directions;
      `test_claude_md_gates.py` (the citation gate, which reads `CHANGELOG.md` paths) passes. Full
      local suite running at push time — result posted in the thread below rather than ticked here.
      No new test file, so `run_tests.py` `TESTS` is unchanged.
- [x] Web: no TypeScript changed, so tsc/vitest/build were not re-run
- [x] `CHANGELOG.md` entry added under `## Unreleased`, plus the correction to #540's figure
- [x] No secrets; no competitor names

---

## Summary by CodeRabbit

- **Documentation**
  - Updated reachability documentation to record 70 routes affected by a measured blind spot among 947 routes.
  - Clarified that the recorded set is not an allowlist and that route triage remains separate work.
  - Superseded the previous estimate of 26 candidates with the corrected measurement.

- **Tests**
  - Added coverage for blind-spot identification and positive and negative control cases.
  - Added checks to detect newly affected routes and ensure unexpected changes fail validation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,19 @@ with a name. Triage is deliberately *not* done here — no syntactic probe can c
 the leaf being vouched is the whole problem (a leaf-based search returns 685 hits for `projects` and
 44 for `status`). Each needs a read, and that is its own work.
 
+**Two review findings, both real, both applied.** The derivation matched its probe against all
+code; `leaf_is_called` directly above it matches inside STRING LITERALS, for the documented reason
+that a bare substring test cannot tell a route from a token containing it — so a probe appearing in
+an identifier could drop a route from the set, and since the ratchet fails only on *additions*, that
+removal would never be reported. *This file's own recurring defect, one level down, inside the
+derivation written to measure it.* The differential was **measured, not assumed: 0 routes today**,
+both controls unmoved — so it is preventive rather than corrective, and applied because it closes the
+silent-shrink path for nothing. Second: a route *leaving* the frozen set only printed. It now fails,
+on `KNOWN_UNCALLED`'s stated reasoning — an entry leaves either because it gained a caller, or
+because it was renamed, and then the frozen name is a stale string pre-authorising whatever reuses
+that path. *"Both need a human; neither should print and pass."* The weaker precedent in the same
+file was the wrong one to copy.
+
 **This took four measurements to get right** — 377 and 35 in a previous gate, then 26 and 70 here —
 and the pattern is identical each time: *the checker did not fail, it answered.* So the derivation
 carries two live controls it must separate: `/bsdd/search`, wired an hour earlier, must fall out;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,33 @@ meaning anything as a heading and the file read as 34 pending releases rather th
 titles are unchanged and now sit at `###` beneath this, in the same order; no text was edited,
 added or dropped in the fold.
 
+### The reachability gate's blind spot was 70 routes wide, and only two were written down
+
+`LEAF_COLLISION_DARK` is a **triaged** list — routes somebody read, confirmed dark, and recorded. It
+was never the population, and the note above it described it as though it were. The population is
+every route the reachability rule cannot see *at all*: leaf long enough to check, leaf vouched by an
+ordinary English or HTML word (so the rule stays silent), and no caller that builds the path.
+Deriving that instead of listing it gives **70 of 947 routes**.
+
+`services/api/test_route_reachability.py` now derives and freezes that set as `LEAF_COLLISION_BLIND`,
+the way `KNOWN_UNCALLED` already freezes the routes the rule *can* see. It is a **record, not an
+allowlist**: 70 is not a target, and most of the set is probably correct. `/scim/v2/Users` is for an
+identity provider and `/bcf/2.1/projects` for external BIM tools — neither should ever have a web
+caller. Some are false positives of the probe itself: `openDrawing("plan.dxf?…")` builds
+`/projects/{pid}/drawings/plan.dxf` through a helper, so the literal never appears.
+
+**The point is not the number, it is that the number exists.** Before this, the blind spot had no
+measured size and nothing could report a new member; now it can only shrink, and route 71 arrives
+with a name. Triage is deliberately *not* done here — no syntactic probe can classify these, because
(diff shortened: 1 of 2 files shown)


ibuilder:
**Local backend suite: 698/698 suites passed, `run_tests exit: 0`** (3 parallel, 1162s wall).

This closes the checklist item left deliberately unticked at push time — the suite was still running, so the checklist said so rather than claiming a pass. Read from the runner's own exit line, not through a pipe; 0 `FAIL` lines across the run.

`test_route_reachability.py` — the only code file this PR changes — passes both inside the suite and standalone, and was mutated in both directions:

| mutation | result |
|---|---|
| drop `/scim/v2/Users` from the frozen set, simulating a 71st route arriving | red, naming the route |
| blind the derivation so it returns nothing | red on the **controls**, rather than a silently clean tree |

The controls are both live routes, so neither can pass by naming something that has stopped existing: `/bsdd/search` — wired an hour ago in #540 — must fall *out* of the set, and `/scim/v2/Users`, callerless by design, must stay *in*. That pairing is what makes "70 frozen" a measurement rather than an assertion.

No new test file, so `run_tests.py` `TESTS` is unchanged at 695 entries; 695 + the 3 `DATA_TESTS` from `services/data` is the 698 above.

---

ibuilder:
Both nitpicks verified against the code and **both applied** in `b511bce3`. The first is a better catch than "nitpick" suggests.

### 1. Probe matched against all code, not string literals — *correct, and it is this file's own defect*

`leaf_is_called` matches inside `string_blob(code)` deliberately, for the reason recorded in its docstring: *"A bare substring test cannot tell a route from a word that happens to contain it."* `_blind_spot()` matched `probe in stripped`. Same class of error, one level down, **inside the derivation written to measure that class** — and one-directional exactly as described: it can only shrink the set, and since the ratchet fails on *additions*, a route dropped that way is never reported.

Fixed to `probe in string_blob(stripped)`.

**The differential is 0 routes today**, both controls unmoved — measured, not assumed, and I'd rather say that than imply the fix corrected a live miscount:

| | routes |
|---|---|
| matched against all code | 70 |
| matched against string literals | 70 |
| wrongly excluded by the looser match | **0** |

Mutation confirms it: reverting the fix reds nothing. So it is **preventive, not corrective** — applied because it closes the silent-shrink path for free, on the same argument the string-literal restriction above it was made on.

### 2. Ratchet-down printed instead of failing — correct, and I copied the wrong precedent

The file has both patterns. `CONFIRMED_DARK_SHORT_LEAF` prints and passes; I followed it. `KNOWN_UNCALLED`'s rot check does not, and its comment says why in terms that apply here verbatim: an entry leaves either because the route gained a caller (good news the list must record) or because it was **renamed or deleted** — and then the frozen name is a stale string pre-authorising whatever later reuses that path. *"Both need a human; neither should print and pass."*

Now a `check()`. Mutation: inserting a non-existent route into the frozen set reds and names it, where the print-only version passed silently.

Local verification on `b511bce3`: gate green standalone (all checks, including both new ones), `ruff check ../..` clean tree-wide, `test_claude_md_gates.py` (the citation gate over CHANGELOG paths) exit 0. CHANGELOG records both findings and the measured 0 differential.

---

PR #553, opened 2026-09-14
MARKUP-PROMOTE: two promotes on one markup minted two RFIs, one orphaned

ibuilder:
# What & why

`bim.promote_markup` checked `if m.topic_id: 409` and then claimed the back-link with a plain `m.topic_id = t.id`.

That check reads the **session's own copy**, and `SessionLocal` is `expire_on_commit=False` — so a request that loaded the markup before a concurrent promote committed still sees `None` however long it holds it. Both requests pass the guard, both mint an RFI `Topic`, and the later write wins:

```
markup M, topic_id NULL
  A loads M  ->  topic_id None
  B loads M  ->  topic_id None
  B promotes ->  Topic T1, M.topic_id = T1   201
  A promotes ->  Topic T2, M.topic_id = T2   201
  FINAL: two Topics for one markup. T1 is orphaned — nothing points at it.
```

**A double-click is enough.** No load, no contention window worth the name — this is the failure mode a "promote" button produces on the ordinary path, and both callers are told it worked.

The link is now claimed with a conditional `UPDATE ... WHERE id = ? AND topic_id IS NULL` and a rowcount check, rolling back the flushed Topic on a loss so the loser leaves nothing behind. Under Postgres read-committed the loser blocks on the winner's row lock, re-evaluates the predicate against the committed row and matches nothing; under SQLite the writes serialize to the same effect.

### This is not a new idea, and that is the point

It is exactly what `modules.promote_comment` was given in #434 when **the same defect was fixed there**. This instance sat one module over, unchanged, for months.

It was found by RMW-SWEEP's ORM derivation in #552 — *by deriving the population, not by anyone reading this route.* That is the argument for a derived inventory over a listed one: **an already-solved defect living on elsewhere is the cheapest kind to close and the easiest to never look for.**

### Two things the fix exposed

- **The Core `UPDATE` bypasses the ORM.** With `expire_on_commit=False` the session's `m` kept its null `topic_id`, so `_markup_out(m)` would have reported the markup **unpromoted in the very call that promoted it**. A `db.refresh(m)` before the response closes that.
- **Closing the site made its own ledger entry stale**, and `services/api/test_rmw_sweep.py` reported that *before* the ledger was touched — the entry is removed rather than re-statused, because the conditional UPDATE takes the assignment out of the derived population altogether. *Closing a gap has to force the record to move, or the tree ends up describing a defect nobody can still find.*

The Core population also moved 5 → 6 and stayed clean: the new conditional UPDATE is itself now judged by the gate that judges the rest.

## Tests

`services/api/test_markup.py` drives **two real sessions** through the route: the loser reads a null back-link first, the winner promotes, and the loser must be refused with 409 rather than duplicating. Then it asserts the loser minted no orphan (topic count +1, not +2) and that the markup points at the winner's Topic.

**Mutation-verified**: restoring the plain `m.topic_id = t.id` re-fails that test with *"a stale-read promote must be refused, not duplicated"*.

- [x] Backend: `ruff check` clean over `services/api`
- [x] Backend: `test_markup`, `test_comment_promote`, `test_rmw_sweep`, `test_rbac`, `test_claude_md_gates` pass. Full suite running — result posted before this leaves draft
- [x] Web: no TypeScript changed
- [x] `CHANGELOG.md` entry added; `docs/security/threat-model.md` counts corrected (open 12 → 11, G-12 five → four)

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where simultaneous attempts to promote the same markup could create duplicate RFI topics.
  - Only one promotion now succeeds; competing attempts receive a conflict response.
  - Prevented orphaned topics from being created when a promotion loses a race.
  - Promotion responses now immediately show the correct promoted status.

- **Documentation**
  - Updated security documentation and changelog entries to reflect the resolved concurrency issue and revised open-gap counts.


ibuilder:
Marked ready. Local full suite on `6ea1bf40`: **704/704**, 1,257s, 0 test databases unaccounted for.

CI at this head: web typecheck + test + build ✅, alembic upgrade + runtime-parity on Postgres ✅, CodeQL ✅, python analysis ✅, container scope ✅. API test gate still running.

Two things worth a reviewer's attention over the rest of the diff:

**The `db.refresh(m)` is not incidental.** The conditional `UPDATE` is Core, so it bypasses the ORM, and `SessionLocal` is `expire_on_commit=False` — without the refresh the response reports `topic_id: null` for the markup it just promoted. That is the same class of defect as the race itself (a response describing a different row than the one written), introduced by the fix for it, which is why it is called out in the code rather than left as a one-liner.

**The ledger entry was removed, not re-statused.** The conditional UPDATE takes `m.topic_id = t.id` out of the derived read-modify-write population entirely, so an entry describing it would outlive its site — `test_rmw_sweep.py` reported exactly that before the ledger was touched, which is the direction that gate exists for.

---

ibuilder:
Real, and mine. Fixed in `d21a1e21`.

The G-12 entry now says `bim.promote_markup` is closed and its closing sentence still called it the one worth doing first — I updated the status and left the ranking sitting underneath it. A reader following the backlog would have gone to fix a site the same paragraph had just told them was done.

**The first replacement was worse, and that is the part worth recording rather than quietly overwriting.** I ranked `drawingset.revise_sheet` first on the grounds that a lost sheet revision is the most visible of the four and that it has a single writer. Reading the site disproved both inside a minute: the read-modify-write is `m.data = d2` on each **markup** the revision tags `carried_from` — not on the drawing record at all — and that blob is also written by the markup save and bulk-replace routes. So the severity was mischaracterised and the "single writer" was false.

*A priority claim is a factual claim about blast radius*, and this file had then carried two wrong ones about the same list in a row — the stale one you found, and its replacement. So the four remaining sites are now deliberately **not** ranked, with the wrong claim recorded beside them so the next reader knows the ranking has to come from reading the sites rather than from this paragraph.

`test_claude_md_gates` passes (the threat model is one of the docs whose backticked citations it resolves).

---

